PR #16, opened 2019-10-25
Support for LMDB flags

### LINKED ISSUE #15: Is it possible to specify LMDB flags such as MDB_NOSYNC?
Looks like flags are not yet supported. Did I miss them?

Thanks for this nice wrapper.

Laurent

lquerel:
Added the support for the LMDB flags we discussed earlier.

Closes #15.

Laurent


Kerollmops:
Thank you for the contribution but isn’t a flush/sync function is necessary to force LMDB to persist data on disk if flags like `NoSync/NoMetaSync` is enabled?

src/flags.rs
```diff
@@ -0,0 +1,18 @@
...
+    MdbNosSubDir = 0x4000,
+    MdbNoSync = 0x10000,
+    MdbRdOnly = 0x20000,
+    MdbNoMetaSync = 0x40000,
+    MdbWriteMap = 0x80000,
+    MdbMapAsync = 0x100000,
+    MdbNoTls = 0x200000,
+    MdbNoLock = 0x400000,
+    MdbNoRdAhead = 0x800000,
+    MdbNoMemInit = 0x1000000,
```
Kerollmops: I would prefer that you use the [`lmdb-rkv-sys` already extracted flags values](https://docs.rs/lmdb-rkv-sys/0.9.5/lmdb_sys/constant.MDB_FIXEDMAP.html) instead of redefining those by hand.

lquerel:
Thanks for your review, below a description of the modifications for this new PR:
- Added a force_sync method to the Env struct. 
- Reused the constant defined in lmdb_sys.
- Removed the dependency to enumflag2 because it was not compatible with the reference to lmdb_sys constants (enumflags2 only supports literals). However I kept a C-style enum for clarity and soundness.

Kerollmops:
Thank you for this pull request. I will squash your commits.

Sorry that I didn't provide a contributing guide, next time would it be possible to rebase your commits instead of merging master into your branch? 😊

PR #43, opened 2020-05-20
Add parent_write_txn parameter to create db calls

sphw:
This PR adds two to function calls that allow the user to pass in an existing write_txn to the create_database functions.

I believe this should still be safe since we use a nested transaction and still commit it before releasing the lock

### DIFF
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -274,21 +274,49 @@ impl Env {
         KC: 'static,
         DC: 'static,
     {
+        let mut parent_wtxn = self.write_txn()?;
+        let db = self.create_database_with_txn(name, &mut parent_wtxn)?;
+        parent_wtxn.commit()?;
+        Ok(db)
+    }
+
+    pub fn create_database_with_txn<KC, DC>(
+        &self,
+        name: Option<&str>,
+        parent_wtxn: &mut RwTxn,
+    ) -> Result<Database<KC, DC>>
+    where
+        KC: 'static,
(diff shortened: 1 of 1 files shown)


Kerollmops:
Hey,

Thank you @sphw for this PR.
It would be better if the `raw_create_database` function always accepts a transaction, `create_database` and `create_poly_database` then start a write transaction and give it to the `create_database_with_txn` and `create_poly_database_with_txn` respectively.

What do you think?

Also next time could avoid formatting the code please, it would be easier for me to review the changes, or do it in another commit, no worries :)

sphw:
Yeah that makes more sense. I'll modify it to do that.

I'll try and remove the formatting as well, my editor does it automatically

sphw:
@Kerollmops I have updated the PR with the changes. One minor downside to this method is that we create two transactions for `create_database` instead of one. This will probably have a minor affect on performance. I suspect it won't be a problem in the real world since database creation won't happen very often.

Kerollmops:
Hum... Yeah, you are right, I didn't thought about this. This is like you said, not that impactfull.

Could you rebase on master, please? I have merged another PR, I will merge just after you rebased :)

sphw:
> Hum... Yeah, you are right, I didn't thought about this. This is like you said, not that impactfull.
> 
> Could you rebase on master, please? I have merged another PR, I will merge just after you rebased :)

Done

Kerollmops:
Thank you!

PR #52, opened 2020-06-26
compose byte encode decode with pair

MarinPostma:
This is a POC of how we could compose types to that implement `ByteEncode` and `ByteDecode`. It turns out that algrebraic data types preserve these properties, so an implemention of `ByteEncode` and `ByteDecode` could easily be derived for struct and enums made of types implementing those types.

### DIFF
--- a/heed-traits/src/lib.rs
+++ b/heed-traits/src/lib.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::mem;
 
 pub trait BytesEncode<'a> {
     type EItem: ?Sized + 'a;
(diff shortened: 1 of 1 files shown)


heed-traits/src/lib.rs
```diff
@@ -11,3 +12,44 @@ pub trait BytesDecode<'a> {
...
+
+    fn bytes_encode(item: &'a Self::EItem) -> Option<Cow<[u8]>> {
+        let first_bytes = A::bytes_encode(&item.0)?;
+        let second_bytes = B::bytes_encode(&item.1)?;
+        let mut buffer = Vec::with_capacity(
+            mem::size_of::<usize>()
+            + &first_bytes.len()
+            + &second_bytes.len());
+        // signify the length of the first item
+        buffer.extend_from_slice(&first_bytes.len().to_be_bytes());
```
Kerollmops, line 32: On 32bits platforms this is an u32 and on 64bits platforms this is an u64, it is not consistent, you should not do that but rather cast it into an u64.

heed-traits/src/lib.rs
```diff
@@ -11,3 +12,44 @@ pub trait BytesDecode<'a> {
...
+    A::EItem: Sized,
+    B: BytesEncode<'a>,
+{
+    type EItem = (A::EItem, B::EItem);
+
+    fn bytes_encode(item: &'a Self::EItem) -> Option<Cow<[u8]>> {
+        let first_bytes = A::bytes_encode(&item.0)?;
+        let second_bytes = B::bytes_encode(&item.1)?;
+        let mut buffer = Vec::with_capacity(
+            mem::size_of::<usize>()
```
Kerollmops, line 28: Same here, never use an usize when serializing things and storing length.

heed-traits/src/lib.rs
```diff
@@ -11,3 +12,44 @@ pub trait BytesDecode<'a> {
...
+impl<'a, A, B> BytesDecode<'a> for (A, B)
+where
+    A: BytesDecode<'a>,
+    B: BytesDecode<'a>,
+{
+    type DItem = (A::DItem, B::DItem);
+
+    fn bytes_decode(bytes: &'a [u8]) -> Option<Self::DItem> {
+        let mut size_buf = [0; 8];
+        size_buf.copy_from_slice(bytes.get(0..8)?);
```
Kerollmops, line 48: You used `size_of<usize>` above but here you use 8, hum....

MarinPostma: good point, gonna fix this

PR #63, opened 2020-10-27
Introduce a real prefix iterator

Kerollmops:
Instead of using the `Ro/RwRange` iterator I introduce a real prefix iterator that works correctly in some specific cases where the bytes is `0xFF` and the end range key wasn't correctly generated (because it uses the `advance_key` function) giving errors in some specific cases.


Kerollmops:
The error is always the same, this related to windows that fails to compile with mdbx. So merging tomorrow maybe.

PR #64, opened 2020-10-27
Close environment safely

Kerollmops:
The current heed version don't expose any way to close an environment because many requierements are needed by either lmdb or mdbx to safely do that. Once closed, no active transaction must be used on any database, in short: all databases and transactions must be closed or aborted, if this is not the case, a `SIGSEGV` will be thrown.

I found a pretty way to expose a close api to the environment, [I added a lifetime to the transactions, the lifetime of the environment they were made from](https://github.com/Kerollmops/heed/pull/64/files?file-filters%5B%5D=.rs#diff-c40afad6bbd223832a2f86c38702313beb78366bd30bbfcd6d014e8cc651dafeR9-R13), this indicates that no transaction can live longer than the environment and therefore when an environment is dropped, [the last reference (because it is `Arc`ed) is dropped, the environment can be safely closed](https://github.com/Kerollmops/heed/pull/64/files?file-filters%5B%5D=.rs#diff-ac577a0bca84e6fbce36b3a413107331b37c74809219009e740896600a1d97f7R229-R242).

The close api is a `prepare_for_closing` self consuming method on the `Env` struct which returns a `EnvClosingEvent` that allows the user to wait for the environment to effectively close.

```rust
let env = EnvOpenOptions::new().open("my-env.mdb")?;
let closing_event = env.prepare_for_closing();
closing_event.wait(); // once it unblocks the thread, the env has been closed.

// another thread can also use the env_closing_event function
if let Some(closing_event) = env_closing_event("my-env.mdb") {
    closing_event.wait();
}
```

I also added a new safety guard to the transactions and databases to make sure that a transaction is always used on a database which comes from the same environment, I store the environment pointer (as an `usize`) in the `Database` this way [we are able to assert that the transaction environment pointer (which is referenced in the transaction) is equal to the one in the Database](https://github.com/Kerollmops/heed/pull/64/files#diff-be1d9b2d86fd3893a95f16a687e9e2c5d7e0c609f1840319437f9baeebd736f0R450-R454).

### DIFF
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -493,3 +493,54 @@ impl EnvClosingEvent {
         self.0.wait_timeout(timeout);
     }
 }
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn close_env() {
+        use std::{fs, thread};
+        use std::time::Duration;
+        use std::path::Path;
+        use crate::EnvOpenOptions;
+        use crate::types::*;
+        use crate::env_closing_event;
+
+        fs::create_dir_all(Path::new("target").join("close-env.mdb")).unwrap();
+        let env = EnvOpenOptions::new()
+            .map_size(10 * 1024 * 1024) // 10MB
+            .max_dbs(30)
+            .open(Path::new("target").join("close-env.mdb")).unwrap();
+
+        // Force a thread to keep the env for 1 second.
+        let env_cloned = env.clone();
+        thread::spawn(move || {
+            let _env = env_cloned;
+            thread::sleep(Duration::from_secs(1));
+        });
+
+        let db = env.create_database::<Str, Str>(None).unwrap();
+
+        // Create an ordered list of keys...
+        let mut wtxn = env.write_txn().unwrap();
+        db.put(&mut wtxn, "hello", "hello").unwrap();
+        db.put(&mut wtxn, "world", "world").unwrap();
+
+        // Lets check that we can prefix_iter on that sequence with the key "255".
+        let mut iter = db.iter(&wtxn).unwrap();
+        assert_eq!(iter.next().transpose().unwrap(), Some(("hello", "hello")));
+        assert_eq!(iter.next().transpose().unwrap(), Some(("world", "world")));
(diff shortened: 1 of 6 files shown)


heed/src/env.rs
```diff
@@ -210,13 +217,27 @@ unsafe impl Send for EnvInner {}
 
 unsafe impl Sync for EnvInner {}
 
+impl Drop for EnvInner {
+    fn drop(&mut self) {
+        // We lock the environment map to ensure that no one tries to
+        // open this environment at the same time as we are closing it.
+        let _lock = OPENED_ENV.lock().unwrap();
```
Kerollmops, line 224: This will obviously trigger a dead-lock as will already locked the `OPENED_ENV` mutex to be able to drop the last reference of the env, we could use a `parking_lot` mutex as it is reentrant, when two locks are acquiered on the same thread it works fine.

Kerollmops: Or we can maybe replace the `Env` in the `OPENED_ENV` `HashMap` by a `None` to indicate that it is being dropped right now and acquire the lock again in the `Env::drop` method and remove the entire entry. This way if another thread tries to open the environment when we are trying to `close` it, it must be inform that the env is currently in a closing operation and must not be opened, an error must be returned.

heed/src/env.rs
```diff
@@ -196,6 +194,15 @@ impl EnvOpenOptions {
...
     }
+
+    pub fn close<P: AsRef<Path>>(path: P) -> Result<()> {
+        let path = canonicalize_path(path.as_ref())?;
+
+        let mut lock = OPENED_ENV.lock().unwrap();
+        let _env = lock.remove(&path);
+
+        Ok(())
+    }
```
Kerollmops, line 205: This close method should not be on `EnvOpenOptions`, it doesn't makes sens, I should move that to the `Env` struct itself.

Kerollmops:
I used [the synchronoise](https://docs.rs/synchronoise/1.0.0/synchronoise/event/struct.SignalEvent.html) library instead of a `Mutex` x `CondVar` impl as it seemed safer and better than something that I would have spent time on and that could highly trigger dead-locks.

PR #77, opened 2020-11-19
bump deps

MarinPostma:


Kerollmops:
Thank you, can you check if there are more dependencies that can be bumped?
This way I will accept and publish a version with all these dependencies bumped!

Kerollmops (approved):
Thank you very much

PR #79, opened 2020-11-24
Efficient Iterator last method

Kerollmops:
Improve the `Iterator::last` method on `Ro/RwIter`  and `Ro/RwRange` iterator structs.

### DIFF
--- a/heed/src/cursor.rs
+++ b/heed/src/cursor.rs
@@ -22,6 +22,31 @@ impl<'txn> RoCursor<'txn> {
         })
     }
 
+    pub fn current(&mut self) -> Result<Option<(&'txn [u8], &'txn [u8])>> {
+        let mut key_val = mem::MaybeUninit::uninit();
+        let mut data_val = mem::MaybeUninit::uninit();
+
+        // Move the cursor on the first database key
+        let result = unsafe {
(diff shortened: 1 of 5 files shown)


Kerollmops:
Hey @placrosse,

> I saw that issue about the double cursor. What I really need is to be able to select the last valid key from a range, without iterating through everything. I didn't see how it was possible in the current API.

I know [that it is long ago](https://github.com/Kerollmops/heed/issues/30#issuecomment-568061207) and you have probably found a workaround since then but I just improved the `Iterator::last` methods for the `Ro/RwIter`, `Ro/RwRange` and `Ro/RwPrefix` iterators.

I also introduced some useful methods on databases to [get `greater/lower_than` a specified key entry](https://github.com/Kerollmops/heed/pull/72) and to [get `greater/lower_than_or_equal_to` a specified key entry](https://github.com/Kerollmops/heed/pull/73).

So if you want you can test those with your project. In a future PR [I will introduce reverse iterators for ranges, prefixes, and simple iterators](https://github.com/Kerollmops/heed/issues/30#issuecomment-569541501), but will not implement [`DoubleEndedIterator`](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html) on them.

PR #80, opened 2020-11-25
Introduce reverse iterators

### LINKED ISSUE #30: Trait std::iter::DoubleEndedIterator
Since a LMDB cursor is capable of quickly going to last, and getting the previous entry, it should support the DoubleEndedIterator, allowing last to first iteration of a range.

Kerollmops:
This PR will fix #30 in part as it exposes a lot of new ways to iterate over the database entries in reverse order. However, [for technical reasons](https://github.com/Kerollmops/heed/issues/30#issuecomment-569408011), it doesn't implement [the DoubleEndedIterator trait](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html) on the currently available structs.

In short, this PR introduces the `RoRevIter`, `RwRevIter`, `RoRevRange`, `RwRevRange`, `RoRevPrefix`, and the `RwRevPrefix` iterators that allows the user to iterate on the database entries but in reverse order. All of these iterators efficiently implement [`Iterator::last`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.last).

## Example

```rust
let db: Database<OwnedType<BEI32>, Str> = env.create_database(Some("iter-i32"))?;

db.put(&mut wtxn, &BEI32::new(42), "i-am-forty-two")?;
db.put(&mut wtxn, &BEI32::new(27), "i-am-twenty-seven")?;
db.put(&mut wtxn, &BEI32::new(13), "i-am-thirteen")?;
db.put(&mut wtxn, &BEI32::new(521), "i-am-five-hundred-and-twenty-one")?;

let range = BEI32::new(27)..=BEI32::new(43);
let mut iter = db.rev_range(&wtxn, &range)?;
assert_eq!(iter.next().transpose()?, Some((BEI32::new(42), "i-am-forty-two")));
assert_eq!(iter.next().transpose()?, Some((BEI32::new(27), "i-am-twenty-seven")));
assert_eq!(iter.next().transpose()?, None);
```


Kerollmops:
Hey @placrosse,

I would like you to take a look at this PR, as it seems to fix some of the issues you were facing a while ago in #30. If it doesn't fit your needs (especially the `DoubleEndedIterator` story) I would be glad we thought about this, as maybe it can be implemented efficiently and without much overhead (without having to store the current key for both ways).

PR #89, opened 2020-12-31
Byte decode error

### LINKED ISSUE #88: Support custom encoding/decoding errors
It would be better and easier to debug a program if the `BytesEncoding` and `BytesDecoding` traits could return any error type.

To do so we need to modify the `Error` enum and more specifically the `Encoding` and `Decoding` variants to wrap a `Box<dyn Error>`.

MarinPostma:
close #88

### DIFF
--- a/heed-types/src/cow_slice.rs
+++ b/heed-types/src/cow_slice.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::error::Error;
 use std::{mem, ptr};
 
 use crate::aligned_to;
@@ -28,8 +29,8 @@ where
 {
     type EItem = [T];
 
-    fn bytes_encode(item: &'a Self::EItem) -> Option<Cow<[u8]>> {
-        Some(Cow::Borrowed(<[T] as AsBytes>::as_bytes(item)))
+    fn bytes_encode(item: &'a Self::EItem) -> Result<Cow<[u8]>, Box<dyn Error>> {
+        Ok(Cow::Borrowed(<[T] as AsBytes>::as_bytes(item)))
     }
 }
 
@@ -39,9 +40,9 @@ where
 {
     type DItem = Cow<'a, [T]>;
 
-    fn bytes_decode(bytes: &'a [u8]) -> Option<Self::DItem> {
+    fn bytes_decode(bytes: &'a [u8]) -> Result<Self::DItem, Box<dyn Error>> {
         match LayoutVerified::<_, [T]>::new_slice(bytes) {
-            Some(layout) => Some(Cow::Borrowed(layout.into_slice())),
+            Some(layout) => Ok(Cow::Borrowed(layout.into_slice())),
             None => {
                 let len = bytes.len();
                 let elem_size = mem::size_of::<T>();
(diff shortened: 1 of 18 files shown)


heed-types/src/cow_slice.rs
```diff
@@ -58,10 +59,10 @@ where
                         vec.set_len(elems);
                     }
 
-                    return Some(Cow::Owned(vec));
+                    return Ok(Cow::Owned(vec));
                 }
 
-                None
+                Err("The provided bytes do not satisfy the alignment requirements.")?
```
Kerollmops, line 65: When I switch from zerocopy to bytemuck I will change this error string into [a nicer error, more precise](https://docs.rs/bytemuck/1.4.1/bytemuck/enum.PodCastError.html)!

heed-types/src/lib.rs
```diff
@@ -71,8 +71,8 @@ pub struct DecodeIgnore;
 impl heed_traits::BytesDecode<'_> for DecodeIgnore {
     type DItem = ();
 
-    fn bytes_decode(_bytes: &[u8]) -> Option<Self::DItem> {
-        Some(())
+    fn bytes_decode(_bytes: &[u8]) -> Result<Self::DItem, Box<dyn std::error::Error>> {
```
Kerollmops, line 74:
```suggestion
    fn bytes_decode(_bytes: &[u8]) -> Result<Self::DItem, Box<dyn Error>> {
```

By importing the Error trait into the scope.

heed/src/db/polymorph.rs
```diff
@@ -278,7 +278,7 @@ impl PolyDatabase {
     {
         assert_eq!(self.env_ident, txn.env.env_mut_ptr() as usize);
 
-        let key_bytes: Cow<[u8]> = KC::bytes_encode(&key).ok_or(Error::Encoding)?;
+        let key_bytes: Cow<[u8]> = KC::bytes_encode(&key).map_err(|e| Error::Encoding(e))?;
```
Kerollmops, line 281:
```suggestion
        let key_bytes: Cow<[u8]> = KC::bytes_encode(&key).map_err(Error::Encoding)?;
```

heed/src/lib.rs
```diff
@@ -80,8 +80,8 @@ use std::{error, fmt, io, result};
 pub enum Error {
     Io(io::Error),
     Mdb(MdbError),
-    Encoding,
-    Decoding,
+    Encoding(Box<dyn std::error::Error>),
+    Decoding(Box<dyn std::error::Error>),
```
Kerollmops, line 84:
```suggestion
    Encoding(Box<dyn Error>),
    Decoding(Box<dyn Error>),
```

Kerollmops (changes requested):
Thank you, I tried but haven't finished yet. So, you take the lead and I will accept your PR!

----

I am wondering if introducing a custom Result type named `CodecResult`, could be better.
```
type CodecResult<T> = Result<T, Box<dyn Error>>;
```

MarinPostma:
> I am wondering if introducing a custom public Result type named CodecResult, could be better.

Why not exposing an anyhow Result instead?

Kerollmops:
> Why not exposing an anyhow Result instead?

I don't want to bring one more dependency in this crate, don't want to force users to use anyhow, and anyhow has been design to be used by binary crates.

Exposing a type alias will work just fine :)

PR #90, opened 2021-01-07
Custom key comparison functions

### LINKED ISSUE #86: Support for custom key comparison function
LMDB in its great design supports a lot of features, one that is quite interesting though is the fact that it allows users [to define their own comparison function](http://www.lmdb.tech/doc/group__mdb.html#ga68e47ffcf72eceec553c72b1784ee0fe). By default, when no custom function is defined, the keys are compared lexically, with shorter keys collating before longer keys.

If we want to support this, a new design must be found. This comparison function must be Rust idiomatic in the sense that it must get decoded values (not the raw bytes) as parameters and return [an `Ordering` struct](https://doc.rust-lang.org/std/cmp/enum.Ordering.html). It would be preferable if this function doesn't impact the `Database` type, in the sense that it would be preferable to use an `Option<fn>` instead of an `Option<F> where F: Fn()`, this way it would make [the `Database` as easily `Clonable/Copyable`](https://doc.rust-lang.org/std/primitive.fn.html#traits) as before.

Kerollmops:
This PR adds some new method for database opening to let the user specify a custom key comparison function if necessary and fixes #86.

- [x] Expose an API to create/open databases with a custom key comparison function.
- [x] Ensure that a database correctly uses a comparison function if it has been open with one.
- [x] Expose a safe Rust API for the user (instead of an `extern "C"` one).
- [x] Move this `CustomKeyCmp` trait into the `heed-trait` crate.
- [x] Ask for a typed function (using the user-defined key codec) instead of a function that takes byte slices.
- [x] Change back the `CustomKeyCmp` trait inputs to be byte slices again, this way we avoid the decoding overheads.
- [x] Make it compile with MDBX but don't let users use that feature as [custom key comparator are deprecated](https://github.com/erthink/libmdbx/blob/adc720816970335d6b7ef85e8a08ddcb336f0dd0/mdbx.h#L3280-L3295).

### Expose a Rust API instead of the LMDB required function signature

The current exposed API for the custom key comparison function is an ugly `extern "C"` function, I would like to let the user give a real Rust function instead. It is ugly in the sense that I don't want to expose the `MDB_val` type (a struct with start and length fields) and therefore the `from_val` function (transform an `MDB_val` into a Rust slice).

```rust
// What the API user must define itself, we give that to LMDB directly.
extern "C" fn my_custom_key_cmp(a: *const MDB_val, b: *const MDB_val) -> i32;

// What I would like to ask the user to give and wrap.
fn my_custom_key_cmp(a: &[u8], b: &[u8]) -> Ordering;

// The wrapper that could be used to call the user defined function,
// transforming a C call into a Rust call, invisible to the API user.
extern "C" fn wrapper(a: *const MDB_val, b: *const MDB_val) -> i32 {
    // transform a and b into Rust slices here using `from_val`.
    // PROBLEM how can I pass the user function here? There is no easy way.
    match (user_cmp_fn)(a, b) {
        Ordering::Less => -1,
        Ordering::Equal => 0,
        Ordering::Greater => 1,
    }
}
```

As it is not quite possible to pass the user-defined function to a wrapper function with [the LMDB required signature](http://www.lmdb.tech/doc/group__mdb.html#gaa8e6e7a6f99bd7142947c48f0c4b970f), I thought that we could maybe be able to "generate" a function with the required signature that is monomorphized to call the user-defined function (no hidden parameter), but is it possible to do that, I am not sure, I tried by using a closure and a [`Box::into_raw` call](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.into_raw), but it didn't work well (`dyn Fn` can't be cast into `extern "C" fn`).

### Ask for a typed key comparison function instead of a function that takes byte slices

The current exposed API asks for a Rust function that takes two bytes slices as arguments and returns an `Ordering`, it is a good enough solution but it would maybe be better to ask for a function that takes decoded types instead, using the database key codec.

```rust
// The trait currently asked the user that is used to compare two keys.
pub trait CustomKeyCmp {
    fn compare(a: &[u8], b: &[u8]) -> Ordering;
}

// The trait that I would like to expose to the user.
pub trait CustomKeyCmp<'t> {
    type Key: 't;
    fn compare(a: Self::Key, b: Self::Key) -> Ordering;
}
```

### DIFF
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -300,6 +300,7 @@ impl Env {
             .map(|db| Database::new(self.env_mut_ptr() as _, db)))
     }
 
+    #[cfg(all(feature = "lmdb", not(feature = "mdbx")))]
     pub fn open_database_with_custom_key_cmp<KC, DC, C>(
         &self,
         name: Option<&str>,
@@ -385,6 +386,7 @@ impl Env {
         Ok(db)
     }
 
+    #[cfg(all(feature = "lmdb", not(feature = "mdbx")))]
     pub fn create_database_with_custom_key_cmp<KC, DC, C>(
         &self,
         name: Option<&str>,
@@ -414,6 +416,7 @@ impl Env {
             .map(|db| Database::new(self.env_mut_ptr() as _, db))
     }
 
+    #[cfg(all(feature = "lmdb", not(feature = "mdbx")))]
     pub fn create_database_with_txn_and_custom_key_cmp<KC, DC, C>(
         &self,
         name: Option<&str>,
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -19,7 +19,6 @@ use crate::flags::Flags;
 use crate::mdb::error::mdb_result;
 use crate::{Database, CustomKeyCmp, Error, Result, RoTxn, RwTxn};
 use crate::mdb::ffi;
-use crate::BytesDecode;
 
 /// The list of opened environments, the value is an optional environment, it is None
 /// when someone asks to close the environment, closing is a two-phase step, to make sure
@@ -263,20 +262,13 @@ pub enum CompactionOption {
 
 /// An helping function that transforms the LMDB types into Rust types (`MDB_val` into slices)
 /// and vice versa, the Rust types into C types (`Ordering` into an integer).
-extern "C" fn custom_key_cmp_wrapper<'a, KC, C, K>(
+extern "C" fn custom_key_cmp_wrapper<C: CustomKeyCmp>(
     a: *const ffi::MDB_val,
     b: *const ffi::MDB_val,
 ) -> i32
-where
-    KC: BytesDecode<'a, DItem = K>,
-    C: CustomKeyCmp<'a, Key = K>,
 {
     let a = unsafe { ffi::from_val(*a) };
     let b = unsafe { ffi::from_val(*b) };
-
-    let a = KC::bytes_decode(a).unwrap();
-    let b = KC::bytes_decode(b).unwrap();
-
     match C::compare(a, b) {
         Ordering::Less => -1,
         Ordering::Equal => 0,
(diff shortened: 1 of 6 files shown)


myfreeweb:
Perfect solution: ask LMDB to change the signature to add a `void *user_data` that would be passed to the comparator.

Hack: pass the rust function via a `thread_local`? :D

Keep in mind: https://matklad.github.io/2020/10/03/fast-thread-locals-in-rust.html

myfreeweb:
> extern "C" fn wrapper<C: CustomKeyCmp>

Yeah, for just static functions (not closures) this is perfect, and if users want to capture any state, they have to do global passing themselves

Kerollmops:
Thank you for the help @myfreeweb,

Indeed it is not easy to accept a user-defined Rust closure as the custom key comparison function. A lot of people helped me on Reddit and some of them were proposing this solution (using a trait as a generic parameter to an `extern "C"` function) therefore generating the appropriate function for LMDB.

> Yeah, for just static functions (not closures) this is perfect, and if users want to capture any state, they have to do global passing themselves

And yes, this can be a good enough solution for those who really need a Rust closure with a context or something.

heed/src/env.rs
```diff
@@ -271,15 +298,34 @@ impl Env {
...
+        name: Option<&str>,
+    ) -> Result<Option<Database<KC, DC>>>
+    where
+        KC: 'static,
+        DC: 'static,
+        C: CustomKeyCmp,
+    {
+        let types = (TypeId::of::<KC>(), TypeId::of::<DC>());
+        Ok(self
+            .raw_open_database::<DummyCustomKeyCmp>(name, types, true)?
```
Kerollmops, line 314: Wrong, We must use `C` here not the `DummyCustomKeyCmp` type.

Kerollmops:
The complexity is more there than what I thought, introducing a custom key comparison function to the database (that is in fact based on the LMDB one) introduces a lot of little behavior like the fact that we can't just [advance and retreat a key by only using the bytes of it](https://github.com/Kerollmops/heed/blob/e7649c2fb4204a29191f9bef1133bf08faeb8c73/heed/src/iter/range.rs#L40-L42) as it is not necessarily the way it works (think about integer in little-endian).

So I am closing this now as I don't want to take more time on that problem for now.

PR #91, opened 2021-01-16
Moving to bytemuck

### LINKED ISSUE #82: Moving from zerocopy to bytemuck
This library currently uses [the zerocopy crate](https://lib.rs/zerocopy) to provide some useful codecs for common types and types which implements the [`AsBytes`](https://docs.rs/zerocopy/0.3.0/zerocopy/trait.AsBytes.html)/[`FromBytes`](https://docs.rs/zerocopy/0.3.0/zerocopy/trait.FromBytes.html) traits (i.e. `u8`, `[u8]`). I find it complex/impossible to contribute to this crate as there is no repository for this crate only as it has been developed for the new fuchsia OS; sources of this crate are mixed with the source of the OS (in a sub-repository at least).

[The bytemuck library](https://lib.rs/bytemuck) seems to be much easier to contribute to; it is hosted on Github, seems to be much more popular than the former (114k downloads by month compared to 15k). It brings a better API, at least for heed, as it can return [information on which kind of problem happens when a cast fails](https://docs.rs/bytemuck/1.4.1/bytemuck/enum.PodCastError.html), it would [simplify some codecs](https://github.com/Kerollmops/heed/blob/3f170c80b6e39af183bcf481d5afeefeec87cf6e/heed-types/src/cow_slice.rs#L42-L68).

I looked into the crate and found that the change, despite breaking, will be easy. I also found out that we can cast slices of different types by using [the `try_cast_slice` function](https://docs.rs/bytemuck/1.4.1/bytemuck/fn.try_cast_slice.html).

I am just not sure how to expose integers as big-endian, ensuring that the global order is preserved. We will maybe need to create codecs [for every integer type like what zerocopy has done](https://docs.rs/zerocopy/0.3.0/zerocopy/byteorder/index.html) using [byteorder maybe](https://lib.rs/byteorder).

Kerollmops:
Fixes #82 by implementing all of the codecs by using the bytemuck crate and copying the code used in zerocopy and made by @joshlf for the integer types (I licensed the code by adding the Fushia_LICENSE at the root of the repository).

This PR will help #89 return appropriate errors [when alignment is wrong, for example](https://docs.rs/bytemuck/1.5.0/bytemuck/enum.PodCastError.html).

### DIFF
--- a/Fushia_LICENSE
+++ b/Fushia_LICENSE
@@ -1,0 +1,24 @@
+Copyright 2019 The Fuchsia Authors.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+   * Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+   * Redistributions in binary form must reproduce the above
+copyright notice, this list of conditions and the following disclaimer
+in the documentation and/or other materials provided with the
+distribution.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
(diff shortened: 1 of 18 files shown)


Keats:
@Kerollmops 
> Fushia_LICENSE

should be Fuchsia_LICENSE

PR #100, opened 2021-01-28
Use mdb_stat for db.len()

### LINKED ISSUE #56: Improve the poly/database length function
We currently create an iterator count the number of entries to returns to compute the number of entries a database stores.

There is a better way that is not `O(n)` which is using [the `num_entries` field of the `Stat` struct](https://github.com/mozilla/lmdb-rs/blob/df1c2f56e3088f097c719c57b9925ab51e26f3f4/lmdb-sys/src/bindings.rs#L182-L183) that is returned by [the `mdb_stat` function](https://github.com/mozilla/lmdb-rs/blob/df1c2f56e3088f097c719c57b9925ab51e26f3f4/lmdb-sys/src/bindings.rs#L882-L895).

Keats:
Closes #56

### DIFF
--- a/heed/src/db/polymorph.rs
+++ b/heed/src/db/polymorph.rs
@@ -722,24 +722,41 @@ impl PolyDatabase {
     /// assert_eq!(ret, 3);
     ///
     /// wtxn.commit()?;
+    ///
     /// # Ok(()) }
     /// ```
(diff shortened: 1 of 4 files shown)


Keats:
The Windows failure seems unrelated

Kerollmops (approved):
Thank you, and no, the Windows failure is unrelated.

PR #116, opened 2021-07-30
MDB_WRITEMAP fix

timokoesters:
Avoid nested transaction for database, as these are not allowed with MDB_WRITEMAP

### DIFF
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -322,23 +322,23 @@ impl Env {
         KC: 'static,
         DC: 'static,
     {
-        let mut parent_wtxn = self.write_txn()?;
-        let db = self.create_database_with_txn(name, &mut parent_wtxn)?;
-        parent_wtxn.commit()?;
+        let mut wtxn = self.write_txn()?;
+        let db = self.create_database_with_txn(name, &mut wtxn)?;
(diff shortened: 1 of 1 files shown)


Kerollmops (changes requested):
Thank you, but
 - maybe we could remove the `create_poly/_database_with_txn` functions entirely.
 - We no more use the `wtxn` in the `raw_create_database` function so you can also remove this parameter.

Or a better alternative would be to create two `raw_create_database/with_txn` private functions:
- one that takes a `wtxn` and uses it.
- one that doesn't use it.

What do you think? Note that it is ok to break things.

timokoesters:
I want to keep this MR minimal for now and get it merged. These breaking changes can happen in a new MR.

Kerollmops:
Closing this PR as it has been [fixed in v0.20.0-alpha.0 which is available on crates.io](https://github.com/meilisearch/heed/blob/c6179eb2f69f57ace165a18e35d8f945becbdeeb/heed/src/env.rs#L521-L561) or you can even use [the `v0.12` branch where it has also been fixed](https://github.com/meilisearch/heed/commit/ba64ce016e939ff1a35cfaa1989dba7057cb2812).

PR #120, opened 2021-11-24
reentrant prepare for closing

MarinPostma:
Make `prepare_for_closing` reentrant by making it return an `Option<EnvClosingEvent>` rather than panicking when called multiple times.

### DIFF
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -463,17 +463,20 @@ impl Env {
         &self.0.path
     }
 
-    /// Returns an `EnvClosingEvent` that can be used to wait for the closing event,
-    /// multiple threads can wait on this event.
+    /// Deregister the environment from the global environment table so that it can be closed later on,
+    /// when dropped.
+    ///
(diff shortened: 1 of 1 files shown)


heed/src/env.rs
```diff
@@ -552,10 +555,10 @@ mod tests {
 
         wtxn.commit().unwrap();
 
-        let signal_event = env.prepare_for_closing();
```
Kerollmops, line 555: I would prefer that we unwrap here, as it is a test.

Kerollmops (changes requested):

Kerollmops (approved):
Thank you, PR is good, merging this now. Unfortunately, it is breaking and therefore will force me to bump the version of the engine to v0.13.

Kerollmops:
The panic in this function can never be triggered in normal usage, it is not possible to call `prepare_for_closing` on a dropped environment as it is the latest environment dropped that will remove the entry from the HashMap and therefore no other environment can be used to call `prepare_for_closing` on it again.

PR #121, opened 2021-11-24
disable rust fmt

MarinPostma:
disable rustfmt for the project. Note that this is still only available with v1.4.38, that is not shipped with the current stable toolchain (it should be with the next, next week)


Kerollmops:
Thank you, merging this NOW!

PR #123, opened 2022-02-28
reopening an env with different options throws error

MarinPostma:
Return an error when trying to re-open an environment with different settings, instead of quietly returning the original environment.

### DIFF
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -24,7 +24,13 @@ use crate::mdb::ffi;
 /// noone tries to open the same environment between these two phases.
 ///
 /// Trying to open a None marked environment returns an error to the user trying to open it.
-static OPENED_ENV: Lazy<RwLock<HashMap<PathBuf, (Option<Env>, Arc<SignalEvent>)>>> = Lazy::new(RwLock::default);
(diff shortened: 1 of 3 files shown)


heed/src/env.rs
```diff
@@ -24,7 +24,13 @@ use crate::mdb::ffi;
 /// noone tries to open the same environment between these two phases.
 ///
 /// Trying to open a None marked environment returns an error to the user trying to open it.
-static OPENED_ENV: Lazy<RwLock<HashMap<PathBuf, (Option<Env>, Arc<SignalEvent>)>>> = Lazy::new(RwLock::default);
+static OPENED_ENV: Lazy<RwLock<HashMap<PathBuf, EnvCacheEntry>>> = Lazy::new(RwLock::default);
+
+struct EnvCacheEntry {
```
Kerollmops, line 29: This is not a cache entry in the sense that it is required for the library to work and to be safely used and that it can't be evicted from this `HashMap` after a certain time of anything other than closing the env itself.

```suggestion
static OPENED_ENV: Lazy<RwLock<HashMap<PathBuf, EnvEntry>>> = Lazy::new(RwLock::default);

struct EnvEntry {
```

Kerollmops (changes requested):
Please fix the CI 😂

Kerollmops (approved):
Thank you very much, merging!

PR #126, opened 2022-03-07
Introduce move_on_key method on cursors

MarinPostma:
implement the `move_on_key` cursor method, to allow more efficient sequential scans.

After giving it some thought, I have also added a `current` method on the iterator. This is because `move_on_key` can land you somewhere else that you actually expect, and it is useful to be able to know where.

### DIFF
--- a/heed/src/iter/iter.rs
+++ b/heed/src/iter/iter.rs
@@ -235,12 +235,27 @@ impl<'txn, KC, DC> RwIter<'txn, KC, DC> {
         self.remap_types::<KC, LazyDecode<DC>>()
     }
 
+    /// Insert a new value in the database using the cursor, and returns whether the value was
+    /// sucessfully written.
+    ///
+    /// On success, the cursor is moved to the newly inserted value.
+    pub unsafe fn put<'a>(&mut self, key: &'a KC::EItem, data: &'a DC::EItem) -> Result<bool>
+    where
+        KC: BytesEncode<'a>,
+        DC: BytesEncode<'a>,
+    {
+        let key = KC::bytes_encode(key).ok_or(Error::Encoding)?;
+        let data = DC::bytes_encode(data).ok_or(Error::Encoding)?;
+
+        self.cursor.put(&key, &data)
+    }
+
     /// Move the iterator to the provided key, and return the value if it exists.
     ///
     /// If the key doesn't exist, the cursor is moved close to where the requested key would have been.
-    pub fn move_on_key(&mut self, key: &'txn KC::EItem) -> Result<Option<DC::DItem>>
+    pub fn move_on_key<'a>(&mut self, key: &'a KC::EItem) -> Result<Option<DC::DItem>>
     where
-        KC: BytesEncode<'txn>,
+        KC: BytesEncode<'a>,
         DC: BytesDecode<'txn>,
     {
         let key = KC::bytes_encode(key).ok_or(Error::Encoding)?;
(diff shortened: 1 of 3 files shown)


heed/src/iter/iter.rs
```diff
@@ -40,6 +40,33 @@ impl<'txn, KC, DC> RoIter<'txn, KC, DC> {
     pub fn lazily_decode_data(self) -> RoIter<'txn, KC, LazyDecode<DC>> {
         self.remap_types::<KC, LazyDecode<DC>>()
     }
+
+    /// Move the iterator to the provided key, and return the value if it exists.
+    ///
+    /// If the key doesn't exist, the cursor is moved close to the where the requested key would have been.
+    pub fn move_on_key(&mut self, key: &'txn KC::EItem) -> Result<Option<DC::DItem>>
```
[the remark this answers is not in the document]
MarinPostma: Move instead of moves is to remain consistent with the rest of the doc, even though I prefer 3rd singular person too

heed/src/iter/iter.rs
```diff
@@ -37,6 +37,33 @@ impl<'txn, KC, DC> RoIter<'txn, KC, DC> {
...
+    }
+
+    /// Return the key/value pair for the current cursor position.
+    pub fn current(&mut self) -> Result<Option<(KC::DItem, DC::DItem)>>
+    where
+        KC: BytesDecode<'txn>,
+        DC: BytesDecode<'txn>,
+    {
+        match self.cursor.current()? {
+            Some((k, v)) => Ok(Some((KC::bytes_decode(k).ok_or(Error::Decoding)?, DC::bytes_decode(v).ok_or(Error::Decoding)?))),
```
Kerollmops, line 63:
```suggestion
            Some((k, v)) => {
                let key = KC::bytes_decode(k).ok_or(Error::Decoding)?;
                let value = DC::bytes_decode(v).ok_or(Error::Decoding)?;
                Ok(Some((key, value)))
            },
```

heed/src/iter/iter.rs
```diff
@@ -204,6 +231,33 @@ impl<'txn, KC, DC> RwIter<'txn, KC, DC> {
...
```
Kerollmops, line 257:
```suggestion
            Some((k, v)) => {
                let key = KC::bytes_decode(k).ok_or(Error::Decoding)?;
                let value = DC::bytes_decode(v).ok_or(Error::Decoding)?;
                Ok(Some((key, value)))
            },
```

Kerollmops (changes requested):
Just ignore my previous review. I think it looks good to me!

heed/src/iter/iter.rs
```diff
@@ -37,6 +37,36 @@ impl<'txn, KC, DC> RoIter<'txn, KC, DC> {
     pub fn lazily_decode_data(self) -> RoIter<'txn, KC, LazyDecode<DC>> {
         self.remap_types::<KC, LazyDecode<DC>>()
     }
+
+    /// Move the iterator to the provided key, and return the value if it exists.
+    ///
+    /// If the key doesn't exist, the cursor is moved close to the where the requested key would have been.
```
Kerollmops, line 43:
```suggestion
    /// If the key doesn't exist, the cursor is moved close to where the requested key would have been.
```

heed/src/iter/iter.rs
```diff
@@ -204,6 +234,36 @@ impl<'txn, KC, DC> RwIter<'txn, KC, DC> {
     pub fn lazily_decode_data(self) -> RwIter<'txn, KC, LazyDecode<DC>> {
         self.remap_types::<KC, LazyDecode<DC>>()
     }
+
+    /// Move the iterator to the provided key, and return the value if it exists.
+    ///
+    /// If the key doesn't exist, the cursor is moved close to the where the requested key would have been.
```
Kerollmops, line 240:
```suggestion
    /// If the key doesn't exist, the cursor is moved close to where the requested key would have been.
```

Kerollmops (changes requested):

PR #131, opened 2022-08-09
add a method to compute the size of an env

irevoire:
Sorry everything got reformated, I was able to select the hunks for the env file but not for the ffi file.
But basically, only `pub use ffi::mdb_env_stat;`, `pub use ffi::mdb_stat;` and `pub use ffi::MDB_stat;` were added

### DIFF
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -277,10 +277,10 @@ pub enum CompactionOption {
 
 impl Env {
     /// Return the size used by all the databases in the environment.
-    pub fn size(&self) -> Result<usize> {
+    pub fn size(&self) -> Result<u64> {
         let compute_size = |stat: lmdb_sys::MDB_stat| {
-            (stat.ms_leaf_pages + stat.ms_branch_pages + stat.ms_overflow_pages)
-                * stat.ms_psize as usize
+            (stat.ms_leaf_pages + stat.ms_branch_pages + stat.ms_overflow_pages) as u64
+                * stat.ms_psize as u64
         };
 
         let mut size = 0;
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -278,17 +278,19 @@ pub enum CompactionOption {
 impl Env {
     /// Return the size used by all the databases in the environment.
     pub fn size(&self) -> Result<usize> {
+        let compute_size = |stat: lmdb_sys::MDB_stat| {
+            (stat.ms_leaf_pages + stat.ms_branch_pages + stat.ms_overflow_pages)
+                * stat.ms_psize as usize
+        };
+
         let mut size = 0;
 
         let mut stat = std::mem::MaybeUninit::uninit();
         unsafe { mdb_result(ffi::mdb_env_stat(self.0.env, stat.as_mut_ptr()))? };
         let stat = unsafe { stat.assume_init() };
-
-        size += (stat.ms_leaf_pages + stat.ms_branch_pages + stat.ms_overflow_pages)
-            * stat.ms_psize as usize;
+        size += compute_size(stat);
 
         let rtxn = self.read_txn()?;
-
         let dbi = self.raw_open_dbi(&rtxn, None, 0)?;
 
         // we don’t want anyone to open an environment while we’re computing the stats
@@ -304,18 +306,13 @@ impl Env {
             }
 
             let key = String::from_utf8(key.to_vec()).unwrap();
-            let dbi = self.raw_open_dbi(&rtxn, Some(&key), 0);
+            if let Ok(dbi) = self.raw_open_dbi(&rtxn, Some(&key), 0) {
+                let mut stat = std::mem::MaybeUninit::uninit();
+                unsafe { mdb_result(ffi::mdb_stat(rtxn.txn, dbi, stat.as_mut_ptr()))? };
+                let stat = unsafe { stat.assume_init() };
 
-            if dbi.is_err() {
-                continue;
+                size += compute_size(stat);
             }
-
-            let mut stat = std::mem::MaybeUninit::uninit();
-            unsafe { mdb_result(ffi::mdb_stat(rtxn.txn, dbi.unwrap(), stat.as_mut_ptr()))? };
-            let stat = unsafe { stat.assume_init() };
-
-            size += (stat.ms_leaf_pages + stat.ms_branch_pages + stat.ms_overflow_pages)
-                * stat.ms_psize as usize;
         }
 
         Ok(size)
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -295,7 +295,7 @@ impl Env {
 
         // we don’t want anyone to open an environment while we’re computing the stats
         // thus we take a lock on the dbi
-        let _lock = self.0.dbi_open_mutex.lock().unwrap();
+        let dbi_open = self.0.dbi_open_mutex.lock().unwrap();
 
         // We’re going to iterate on the unnamed database
         let mut cursor = RoCursor::new(&rtxn, dbi)?;
@@ -312,6 +312,11 @@ impl Env {
                 let stat = unsafe { stat.assume_init() };
 
                 size += compute_size(stat);
+
+                // if the db wasn’t already opened
+                if !dbi_open.contains_key(&dbi) {
+                    unsafe { ffi::mdb_dbi_close(self.0.env, dbi) }
+                }
             }
         }
 
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -1,15 +1,15 @@
 use std::any::TypeId;
 use std::collections::hash_map::{Entry, HashMap};
 use std::ffi::CString;
+#[cfg(windows)]
+use std::ffi::OsStr;
 use std::fs::File;
+#[cfg(unix)]
+use std::os::unix::ffi::OsStrExt;
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, RwLock};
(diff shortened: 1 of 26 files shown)


heed/src/env.rs
```diff
@@ -260,6 +261,57 @@ pub enum CompactionOption {
 }
 
 impl Env {
+    /// Return the size used by all the databases in the environment.
+    pub fn size(&self) -> Result<usize> {
+        let mut size = 0;
+
+        let mut stat = std::mem::MaybeUninit::uninit();
+        unsafe { mdb_result(ffi::mdb_env_stat(self.0.env, stat.as_mut_ptr())) }?;
```
Kerollmops, line 269:
```suggestion
        unsafe { mdb_result(ffi::mdb_env_stat(self.0.env, stat.as_mut_ptr()))? };
```

heed/src/env.rs
```diff
@@ -260,6 +261,57 @@ pub enum CompactionOption {
...
+        let stat = unsafe { stat.assume_init() };
+
+        size += (stat.ms_leaf_pages + stat.ms_branch_pages + stat.ms_overflow_pages)
+            * stat.ms_psize as usize;
```
Kerollmops, line 273: Could we create a _local_ closure to do this computation? instead of duplicating the code two times?

heed/src/env.rs
```diff
@@ -280,28 +332,36 @@ impl Env {
                 .map(|db| PolyDatabase::new(self.env_mut_ptr() as _, db)))
     }
 
-    fn raw_open_database(
+    fn raw_open_dbi(
```
Kerollmops, line 335: Would it be possible to use this function in `raw_create_database` too? And also in all other places where this is useful?

heed/src/mdb/lmdb_ffi.rs
```diff
@@ -1,41 +1,46 @@
...
-pub use ffi::MDB_CREATE as MDB_CREATE;
-pub use ffi::MDB_CURRENT as MDB_CURRENT;
-pub use ffi::MDB_RDONLY as MDB_RDONLY;
-
-pub use ffi::mdb_env_close as mdb_env_close;
+pub use ffi::mdb_filehandle_t;
+pub use ffi::MDB_cursor;
+pub use ffi::MDB_dbi;
+pub use ffi::MDB_env;
+pub use ffi::MDB_envinfo;
```
Kerollmops, line 7: Why this struct has been added?

heed/src/env.rs
```diff
@@ -260,6 +261,57 @@ pub enum CompactionOption {
...
+            * stat.ms_psize as usize;
+
+        let rtxn = self.read_txn()?;
+
+        let dbi = self.raw_open_dbi(&rtxn, None, 0)?;
+
+        // we don’t want anyone to open an environment while we’re computing the stats
+        // thus we take a lock on all mutexes
+        let locks = (
+            OPENED_ENV.write().unwrap(),
```
Kerollmops, line 282: I would prefer that you take a `read` lock here, as other threads that want to read the other envs should be able to do so. Am I right?

Kerollmops: In fact, it is not even required to take a lock on the `OPENED_ENV`, right? As the `Env` that is `&self` is an `Arc<EnvInner>`, the lock is therefore replicated on all the owners of this specific `Env`. You can simply take a lock on the `self.dbi_open_mutex`.

heed/src/env.rs
```diff
@@ -260,6 +261,57 @@ pub enum CompactionOption {
...
+        // We’re going to iterate on the unnamed database
+        let mut cursor = RoCursor::new(&rtxn, dbi)?;
+
+        while let Some((key, _value)) = cursor.move_on_next()? {
+            if key.contains(&0) {
+                continue;
+            }
+
+            let key = String::from_utf8(key.to_vec()).unwrap();
+            let dbi = self.raw_open_dbi(&rtxn, Some(&key), 0);
```
Kerollmops, line 295: We could maybe change the `raw_open_dbi` function signature to directly accept a `&[u8]`, or not? What do you think? Instead of converting it into an UTF8 string.

heed/src/env.rs
```diff
@@ -260,6 +261,57 @@ pub enum CompactionOption {
...
+            if dbi.is_err() {
+                continue;
+            }
+
+            let mut stat = std::mem::MaybeUninit::uninit();
+            unsafe { mdb_result(ffi::mdb_stat(rtxn.txn, dbi.unwrap(), stat.as_mut_ptr()))? };
+            let stat = unsafe { stat.assume_init() };
+
+            size += (stat.ms_leaf_pages + stat.ms_branch_pages + stat.ms_overflow_pages)
+                * stat.ms_psize as usize;
```
Kerollmops, line 306:
```suggestion
            if let Ok(dbi) = self.raw_open_dbi(&rtxn, Some(&key), 0) {
                let mut stat = std::mem::MaybeUninit::uninit();
                unsafe { mdb_result(ffi::mdb_stat(rtxn.txn, dbi.unwrap(), stat.as_mut_ptr()))? };
                let stat = unsafe { stat.assume_init() };

                size += (stat.ms_leaf_pages + stat.ms_branch_pages + stat.ms_overflow_pages)
                    * stat.ms_psize as usize;
            }
```

What do you think about this?

heed/src/env.rs
```diff
@@ -260,6 +261,57 @@ pub enum CompactionOption {
...
+        }
+
+        // once we reach this point we don’t need the lock anymore
+        drop(locks);
```
Kerollmops, line 310: Why drop it here instead of letting Rust do it by itself? There is no performance gain here, right? It is just a little bit more code to maintain.

Kerollmops (changes requested):

heed/src/mdb/lmdb_ffi.rs
irevoire: Oops yes missed it with the reformating of everything! It was for another test

heed/src/env.rs
irevoire: There was is no performance gain, I removed it 👍

heed/src/env.rs
```diff
@@ -260,6 +261,53 @@ pub enum CompactionOption {
 }
 
 impl Env {
+    /// Return the size used by all the databases in the environment.
+    pub fn size(&self) -> Result<usize> {
+        let compute_size = |stat: lmdb_sys::MDB_stat| {
```
Kerollmops, line 266: Not sure that you need to specify the type here. Rust will guess it by itself.

```suggestion
        let compute_size = |stat| {
```

Kerollmops (changes requested):

heed/src/env.rs
```diff
@@ -297,6 +297,11 @@ impl Env {
                 let stat = unsafe { stat.assume_init() };
 
                 size += compute_size(stat);
+
+                // if the db wasn’t already opened
+                if !dbi_open.contains_key(&dbi) {
+                    unsafe { ffi::mdb_dbi_close(self.0.env, dbi) }
+                }
```
Kerollmops, line 304: A better approch would be to not open it a second time if you find it in the `dbi_open` map, right?

Kerollmops (changes requested):

heed/src/env.rs
irevoire: We could but in the `raw_open_database` and `raw_create_database`, we would then need to do this;
```rust
        let name = name.map(|n| CString::new(n).unwrap());
        let name_ptr = match name {
            Some(ref name) => name.as_bytes_with_nul().as_ptr() as *const _,
            None => ptr::null(),
        };
```

I think it’s easier to convert to utf-8 only in the stats 🤔

heed/src/env.rs
irevoire: I don’t know why but _no_
image

heed/src/env.rs
irevoire: but how can I know its dbi without opening it?
I looked at the code in `raw_open_dbi` and noticed the first action was to call mdb_dbi_open so I don’t think there is currently a datastructure doing the association name -> dbi 🤔

Kerollmops: Yeah, you are right, I didn't notice that there was no name associated with the dbi in this map. Yup, it should be good the way you've done it!

heed/src/env.rs
```diff
@@ -278,22 +278,24 @@ pub enum CompactionOption {
 impl Env {
     /// Return the size used by all the databases in the environment.
     pub fn size(&self) -> Result<usize> {
```
Kerollmops, line 280: You should better use an u64 here as the disk space used can be bigger than the addressable space, right?

Kerollmops (changes requested):

heed/src/env.rs
irevoire: I trust you on this one

heed/src/env.rs
```diff
@@ -276,6 +276,53 @@ pub enum CompactionOption {
 }
 
 impl Env {
+    /// Return the size used by all the databases in the environment.
+    pub fn size(&self) -> Result<u64> {
```
Kerollmops, line 280: Could we find a better/more explicit name for this function?
```suggestion
    /// Returns the size used by all the databases on the disk in the environment without taking the free pages into account.
    pub fn real_disk_space_size(&self) -> Result<u64> {
```

Kerollmops (changes requested):

Kerollmops (approved):
Thank you very much for the changes!

PR #133, opened 2022-08-17
provide multiple way to access the size of the environment

irevoire:
# Pull Request

## What does this PR do?
Provide multiple ways to access to the size of the databases.
Also provide a getter on the flags.

### DIFF
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -278,7 +278,7 @@ pub enum CompactionOption {
 impl Env {
     /// The real size used by this environment on disk.
     pub fn real_disk_size(&self) -> Result<u64> {
-        let path = if self.contain_flag(Flags::MdbNoSubDir)? {
+        let path = if self.contains_flag(Flags::MdbNoSubDir)? {
             self.path().to_path_buf()
         } else {
             self.path().join("data.mdb")
@@ -295,7 +295,7 @@ impl Env {
         Ok(or != 0)
     }
 
-    /// Return the raw flags stored the environment was opened with.
+    /// Return the raw flags the environment was opened with.
     pub fn raw_flags(&self) -> Result<u32> {
         let mut flags = std::mem::MaybeUninit::uninit();
         unsafe { mdb_result(ffi::mdb_env_get_flags(self.env_mut_ptr(), flags.as_mut_ptr()))? };
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -290,10 +290,7 @@ impl Env {
     /// Check if a flag was specified when opening the environment.
     pub fn contain_flag(&self, flag: Flags) -> Result<bool> {
         let flags = self.raw_flags()?;
-        println!("{flags:032b} &");
-        println!("{:032b}", flag as u32);
         let or = flags & (flag as u32);
-        println!("{or:032b}");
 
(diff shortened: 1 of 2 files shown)


heed/src/env.rs
```diff
@@ -276,8 +276,48 @@ pub enum CompactionOption {
...
+        Ok(path.metadata()?.len())
+    }
+
+    /// Check if a flag was specified when opening the environment.
+    pub fn contain_flag(&self, flag: Flags) -> Result<bool> {
+        let flags = self.raw_flags()?;
+        println!("{flags:032b} &");
+        println!("{:032b}", flag as u32);
+        let or = flags & (flag as u32);
+        println!("{or:032b}");
```
Kerollmops, line 296:
```suggestion
        let flags = self.raw_flags()?;
        let or = flags & (flag as u32);
```

heed/src/env.rs
```diff
@@ -276,8 +276,48 @@ pub enum CompactionOption {
...
+            self.path().to_path_buf()
+        } else {
+            self.path().join("data.mdb")
+        };
+
+        Ok(path.metadata()?.len())
+    }
+
+    /// Check if a flag was specified when opening the environment.
+    pub fn contain_flag(&self, flag: Flags) -> Result<bool> {
```
Kerollmops, line 291:
```suggestion
    pub fn contains_flag(&self, flag: Flags) -> Result<bool> {
```

heed/src/env.rs
```diff
@@ -276,8 +276,48 @@ pub enum CompactionOption {
...
+        let flags = self.raw_flags()?;
+        println!("{flags:032b} &");
+        println!("{:032b}", flag as u32);
+        let or = flags & (flag as u32);
+        println!("{or:032b}");
+
+        Ok(or != 0)
+    }
+
+    /// Return the raw flags stored the environment was opened with.
```
Kerollmops, line 301: I don't understand this sentence.

heed/src/env.rs
```diff
@@ -276,8 +276,48 @@ pub enum CompactionOption {
...
+    pub fn raw_flags(&self) -> Result<u32> {
+        let mut flags = std::mem::MaybeUninit::uninit();
+        unsafe { mdb_result(ffi::mdb_env_get_flags(self.env_mut_ptr(), flags.as_mut_ptr()))? };
+        let flags = unsafe { flags.assume_init() };
+
+        Ok(flags)
+    }
+
+    /// The map size that was set when configuring the environment.
+    pub fn map_size(&self) -> Result<usize> {
```
Kerollmops, line 311: I would say that the map size should be returned as an u64 instead, am I right?

Kerollmops (changes requested):

heed/src/env.rs
irevoire: Oops yes, changed to;
>  /// Return the raw flags the environment was opened with.

heed/src/env.rs
irevoire: The ffi returns it as an `usize` so I kept it like that.

Kerollmops (approved):
Thank you very much! Will definitely be helpful!
Would it be possible to fix the tests, please?

Kerollmops (approved):
bors merge

PR #143, opened 2023-01-05
Refuse to create new transactions in an environment on which `prepare_for_closing` has been called

dureuill:
# Pull Request

## Related issue
Related to https://github.com/meilisearch/meilisearch/discussions/3280, this contributes to the implementation of point (3)

### User perspective

- Attempting to create a read-only or read-write transaction on a transaction will now return `Error::DatabaseClosing` if there was a prior call to `Environment::prepare_for_closing` from any thread.
- Added `Environment::is_closing` that returns `true` if there was a prior call to `Environment::prepare_for_closing`.

### Implementation perspective

- Add a `closing: AtomicBool` to `InnerEnv`, that is set to `true` before waiting for the lock in `Environment::prepare_for_closing`.
- When creating a transaction, check for that AtomicBool. If `true`, fails with `Error::DatabaseClosing`

## PR checklist
- [ ] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [ ] Have you made sure that the title is accurate and descriptive of the changes?

### DIFF
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -7,6 +7,7 @@ use std::fs::File;
 #[cfg(unix)]
 use std::os::unix::ffi::OsStrExt;
 use std::path::{Path, PathBuf};
+use std::sync::atomic::AtomicBool;
 use std::sync::{Arc, RwLock};
 use std::time::Duration;
 use std::{io, ptr, sync};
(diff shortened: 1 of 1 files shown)


dureuill:
This PR is actually not needed for https://github.com/meilisearch/meilisearch/discussions/3280.

Closing for now (we could decide to put it back later if we need it for something else)

PR #158, opened 2023-01-14
Better support opening an environment with the `MdbNoSubDir` flag

### LINKED ISSUE #145: Env open with `Flags::MdbNoSubDir` flag requires target file to exists
I'm trying to create a DB like "some_dir/custom.mdb". Where "custom.mdb" is a file, not a directory. I use `Flags::MdbNoSubDir` to achieve it. 
The current implementation requires the target file to exist due to the [use of canonicalize](https://github.com/meilisearch/heed/blob/main/heed/src/env.rs#L153), even though underlying `mdb_env_open` works without creating the file.

It works if I create an empty file manually, but it would be great to not touch the DB file explicitly.

Kerollmops:
Fixes #145 by trying to canonicalize only the path without the file if the file doesn't exist yet.

### DIFF
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -178,6 +178,8 @@ impl EnvOpenOptions {
 
     /// Open an environment that will be located at the specified path.
     pub fn open<P: AsRef<Path>>(&self, path: P) -> Result<Env> {
+        let mut lock = OPENED_ENV.write().unwrap();
+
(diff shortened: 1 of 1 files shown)


shedar (approved):
I love how you simplified the idea.

Kerollmops:
Thank you for the review. I am merging, then!

PR #166, opened 2023-02-22
Make the error returned by the BytesDecode/BytesEncode trait customizable

irevoire:
This PR is an attempt at getting rid of the `BoxedError` type we currently use everywhere.
That allows us to use a « real » type when an error arises in the `BytesEncode` or `BytesDecode` traits.

## The new traits

```rust
/// A trait that represents an encoding structure.
pub trait BytesEncode<'a> {
    type EItem: ?Sized + 'a;
    type Err;

    fn bytes_encode(item: &'a Self::EItem) -> Result<Cow<'a, [u8]>, Self::Err>;
}

/// A trait that represents a decoding structure.
pub trait BytesDecode<'a> {
    type DItem: 'a;
    type Err;

    fn bytes_decode(bytes: &'a [u8]) -> Result<Self::DItem, Self::Err>;
}
```

I called the type `Err` to do something similar to the [`FromStr`](https://doc.rust-lang.org/stable/std/str/trait.FromStr.html) trait from the stdlib.

## The new `Error` enum

```rust
/// An error that encapsulates all possible errors in this crate.
#[derive(Debug)]
pub enum Error<E, D> {
    Io(io::Error),
    Mdb(MdbError),
    Encoding(E),
    Decoding(D),
    InvalidDatabaseTyping,
    DatabaseClosing,
    BadOpenOptions {
        /// The options that were used to originaly open this env.
        options: EnvOpenOptions,
        /// The env opened with the original options.
        env: Env,
    },
}
```

I had to make the `Error` enum generic over the decoding and encoding error.

For most functions, that do not add any complexity because we use our `Result` type;
```rust
/// Either a success or an [`Error`].
pub type Result<T, E = Infallible, D = Infallible> = result::Result<T, Error<E, D>>;
```

That set `E` and `D` to `Infallible` by default.

## My issue

For some methods like the following one, we can return an error while encoding the key OR the value + while decoding the key OR the value:
```rust
    pub fn get_lower_than<'a, 'txn, KC, DC>(
        &self,
        txn: &'txn RoTxn,
        key: &'a KC::EItem,
    ) -> Result<Option<(KC::DItem, DC::DItem)>, KC::Err, DC::Err>
    where
        KC: BytesEncode<'a> + BytesDecode<'txn>,
        DC: BytesDecode<'txn>,
    {
        assert_eq_env_db_txn!(self, txn);

        let mut cursor = RoCursor::new(txn, self.dbi)?;
        let key_bytes: Cow<[u8]> = KC::bytes_encode(&key).map_err(Error::Encoding)?;
        cursor.move_on_key_greater_than_or_equal_to(&key_bytes)?;

        match cursor.move_on_prev() {
            Ok(Some((key, data))) => match (KC::bytes_decode(key), DC::bytes_decode(data)) {
                (Ok(key), Ok(data)) => Ok(Some((key, data))),
                (Err(e), _) | (_, Err(e)) => Err(Error::Decoding(e)),
            },
            Ok(None) => Ok(None),
            Err(e) => Err(e),
        }
    }
```

One solution could be adding two other variants to the `Error` enum.

### DIFF
--- a/heed-traits/src/lib.rs
+++ b/heed-traits/src/lib.rs
@@ -1,19 +1,17 @@
 use std::borrow::Cow;
-use std::error::Error as StdError;
-
-/// A boxed `Send + Sync + 'static` error.
-pub type BoxedError = Box<dyn StdError + Send + Sync + 'static>;
 
 /// A trait that represents an encoding structure.
 pub trait BytesEncode<'a> {
     type EItem: ?Sized + 'a;
+    type Err;
 
-    fn bytes_encode(item: &'a Self::EItem) -> Result<Cow<'a, [u8]>, BoxedError>;
+    fn bytes_encode(item: &'a Self::EItem) -> Result<Cow<'a, [u8]>, Self::Err>;
 }
 
 /// A trait that represents a decoding structure.
 pub trait BytesDecode<'a> {
     type DItem: 'a;
+    type Err;
 
-    fn bytes_decode(bytes: &'a [u8]) -> Result<Self::DItem, BoxedError>;
+    fn bytes_decode(bytes: &'a [u8]) -> Result<Self::DItem, Self::Err>;
 }
--- a/heed-types/src/cow_slice.rs
+++ b/heed-types/src/cow_slice.rs
@@ -1,7 +1,7 @@
 use std::borrow::Cow;
 
 use bytemuck::{pod_collect_to_vec, try_cast_slice, AnyBitPattern, NoUninit, PodCastError};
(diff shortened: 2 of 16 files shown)


Kerollmops:
I tried to introduce more variants to the `Error` enum to represent the key encoding/decoding and data encoding/decoding possible errors.

Unfortunately, it's not that easy to convert the generic `Error` types between them. Some functions return a `Result<Infallible, Infallible, Infallible, Infallible>` and it is not easy to convert that into a `Result<KC, KD, DC, DD>`.

PR #169, opened 2023-06-08
Fix the impl of `Sync` for `RoTxn`

Kerollmops:
This PR makes sure that the Sync trait is correctly implemented on the `RoTxn` when the `sync-read-txn` feature is enabled. It also updates the GitHub CI to make sure we do not break it in further versions of the crate.


dureuill (approved):
LGTM

PR #172, opened 2023-06-27
Fix warnings

Kerollmops:
This PR fixes some warnings when either building or testing the crate. Only the `cstr` dev-dependency has been added. And no perf regression should be seen.


dureuill (approved):
Looks good to me 👓 


bors merge

Issue #150, opened 2023-01-11
Improve the documentation by bringing the LMDB caveats into it


Kerollmops:
Bring [the caveats of LMDB](http://www.lmdb.tech/doc/index.html) (from the `mdb.master` branch) into the lib.rs documentation:
   - [ ] MacOs write transaction limits up to 10.
   - [ ] Windows no more than 64 environments at the same time.
   - [ ] Create a table?

PR #174, opened 2023-06-30
RwTxn + RoTxn documentation improvement.

darnuria:
Hello!

Following #150, I feel like it was a good idea to document the possibles errors raisable by: http://www.lmdb.tech/doc/group__mdb.html#gad7ea55da06b77513609efebd44b26920

I didn't included `ENOMEM` since there is no variant yet maybe it's an other whole story.

### DIFF
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -584,6 +584,13 @@ impl Env {
     }
 
     /// Create a transaction with read and write access for use with the environment.
+    ///
+    /// ## LMDB Limitations
+    ///
+    /// Only one [RwTxn] may exist simultaneously in the current environment.
+    /// If another write transaction is initiated, while another write transaction exists
+    /// the thread initiating the new one will wait on a mutex upon completion of the previous
+    /// transaction.
     pub fn write_txn(&self) -> Result<RwTxn> {
         RwTxn::new(self)
     }
@@ -600,6 +607,25 @@ impl Env {
     }
 
     /// Create a transaction with read-only access for use with the environment.
+    ///
+    /// ## LMDB Limitations
+    ///
+    /// It's possible to have multiple read transactions in the same environment
+    /// while there is a write transaction ongoing.
+    ///
+    /// But read transactions prevent reuse of pages freed by newer write transactions,
+    /// thus the database can grow quickly. Write transactions prevent other write transactions,
+    /// since writes are serialized.
+    ///
+    /// So avoid long-lived read transactions.
+    ///
+    /// ## Errors
+    ///
+    /// * [heed::mdb::lmdb_error::Error::Panic]: A fatal error occurred earlier, and the environment must be shut down
+    /// * [heed::mdb::lmdb_error::Error::MapResized]: Another process wrote data beyond this [Env] mapsize and this env
+    /// map must be resized
+    /// * [heed::mdb::lmdb_error::Error::ReadersFull]: a read-only transaction was requested, and the reader lock table is
+    /// full
     pub fn read_txn(&self) -> Result<RoTxn> {
         RoTxn::new(self)
     }
--- a/heed/src/txn.rs
+++ b/heed/src/txn.rs
@@ -6,6 +6,24 @@ use crate::mdb::ffi;
 use crate::{Env, Result};
 
 /// A read-only transaction.
+///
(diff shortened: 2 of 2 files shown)


heed/src/env.rs
```diff
@@ -584,6 +584,13 @@ impl Env {
     }
 
     /// Create a transaction with read and write access for use with the environment.
+    ///
+    /// ## LMDB limitation
+    ///
+    /// Only one [RwTxn] may exist at the same time on the current environnement.
+    /// If an other write transaction is initiated, while an other write transaction exist
+    /// the thread initiating the new one will wait on a mutex upon completion of the previous
+    /// transaction.
```
Kerollmops, line 593:
```suggestion
    /// ## LMDB Limitations
    ///
    /// Only one [RwTxn] may exist simultaneously in the current environment.
    /// If another write transaction is initiated, while another write transaction exists
    /// the thread initiating the new one will wait on a mutex upon completion of the previous
    /// transaction.
```

heed/src/env.rs
```diff
@@ -600,6 +607,25 @@ impl Env {
...
+    ///
+    /// So avoid long lived read transaction.
+    ///
+    /// ## Errors
+    ///
+    /// * [heed::mdb::lmdb_error::Error::Panic]: A fatal error occurred earlier and the environnement must be shut down
+    /// * [heed::mdb::lmdb_error::Error::MapResized]: Another process wrote data beyond this [Env] mapsize and this env
+    /// map must be resized
+    /// * [heed::mdb::lmdb_error::Error::ReadersFull]: a read-only transaction was requested and the reader lock table is
+    /// full
```
Kerollmops, line 628:
```suggestion
    /// ## LMDB Limitations
    ///
    /// It's possible to have multiple read transactions in the same environment
    /// while there is a write transaction ongoing.
    ///
    /// But read transactions prevent reuse of pages freed by newer write transactions,
    /// thus the database can grow quickly. Write transactions prevent other write transactions,
    /// since writes are serialized.
    ///
    /// So avoid long-lived read transactions.
    ///
    /// ## Errors
    ///
    /// * [heed::mdb::lmdb_error::Error::Panic]: A fatal error occurred earlier, and the environment must be shut down
    /// * [heed::mdb::lmdb_error::Error::MapResized]: Another process wrote data beyond this [Env] mapsize and this env
    /// map must be resized
    /// * [heed::mdb::lmdb_error::Error::ReadersFull]: a read-only transaction was requested, and the reader lock table is
    /// full
```

heed/src/txn.rs
```diff
@@ -6,6 +6,13 @@ use crate::mdb::ffi;
 use crate::{Env, Result};
 
 /// A read-only transaction.
+///
+/// ## LMDB Limitation
+///
+/// It's a must to keep read transaction short lived.
+///
+/// Active Read transactions prevent reuse of pages freed
+/// by newer write transactions, thus the database can grow quickly.
```
Kerollmops, line 15:
```suggestion
/// ## LMDB Limitations
///
/// It's a must to keep read transactions short-lived.
///
/// Active Read transactions prevent the reuse of pages freed
/// by newer write transactions, thus the database can grow quickly.
```

heed/src/txn.rs
```diff
@@ -50,6 +57,10 @@ fn abort_txn(txn: *mut ffi::MDB_txn) {
 }
 
 /// A read-write transaction.
+///
+/// Only one [RwTxn] may exist on the same environnement at the same time,
+/// it two exist the new one may wait on a Mutex for [RwTxn::commit] or [RwTxn::abort] of
+/// the first one.
```
Kerollmops, line 63:
```suggestion
/// ## LMDB Limitations
///
/// Only one [RwTxn] may exist in the same environment at the same time,
/// it two exist, the new one may wait on a Mutex for [RwTxn::commit] or [RwTxn::abort] of
/// the first one.
```

Kerollmops (changes requested):
Thank you for this PR 🏂

Kerollmops (changes requested):
Once the other documentation PR lands. Could you please rebase this one on main?

Kerollmops (approved):

Kerollmops:
Thank you very much for the documentation ❤️

PR #178, opened 2023-07-05
Fix clippy warnings

AureliaDolo:
Hello! 

This PR fixes most clippy lints.

For the Safety doc of EnvOpenOptions::flag I retrieved your comment [here](https://github.com/meilisearch/heed/pull/35#discussion_r363101713)

The yet to fix lints are : 
- [x] Safety doc for RwRevIter::append
- [x] Safety doc for Env::copy_to_fd
- [x] [Module inception](https://rust-lang.github.io/rust-clippy/master/index.html#module_inception) for iter modules

I'd happily take your input regarding this issues.

Thanks !


heed/src/env.rs
```diff
@@ -171,8 +177,11 @@ impl EnvOpenOptions {
     /// assert_eq!(ret, Some(5));
     /// # Ok(()) }
     /// ```
+    ///
+    /// # Safety
+    /// It is unsafe to use unsafe LMDB flags such as Mdb_NoSync, Mdb_NoMetaSync, Mdb_NoLock
```
Kerollmops, line 182:
```suggestion
    ///
    /// It is unsafe to use unsafe LMDB flags such as `NoSync`, `NoMetaSync`, or `NoLock`.
```

Kerollmops (changes requested):
Hey @AureliaDolo 👋,

Thank you very much for your PR! It's pretty cool of you.

Could you please rebase on main, as I removed Bors and fixed the CI? After I review this first step of your PR, could you please add a clippy job in the CI?

About the safety messages:

#### Safety doc for `RwRevIter::append`

That's a really important one, good catch! This message must be copy/pasted for every `append` method of the crate, but you must not give anything that comes from the database itself. Always make sure that you clone that on the heap before giving it to this function if it's the case!

#### Safety doc for `Env::copy_to_fd`

I would say that the main reason is that it exposes a `mdb_filehandle_t` with either a `c_int` on UNIX or a `void*` on Windows. Maybe just a note to that that you must make sure the handle is valid before doing anything. WDTY?

#### [Module inception](https://rust-lang.github.io/rust-clippy/master/index.html#module_inception) for iter modules

There are two ways to fix that:
 - Ignore it with an `#[allow(clippy::module_inception)]`.
 - Rename the outer `iter` module into `iterator`.

AureliaDolo:
Thanks for the review ! 

The renaming of the iterator module should not have an impact on users as iterators are re exported.


> Ho :open_mouth: I just noticed that the env::EnvInfo struct is not reexported to the lib.rs file, could you fix it here, please? I don't know why the compiler/doc compiler said nothing.

I think it's because you may have wanted to use EnvInfo in other modules, but not expose it to the great outside

Kerollmops (approved):

Kerollmops:
Looks perfect to me. Thank you very much! Nice job!

PR #179, opened 2023-07-09
Check env with device and Inode instead of paths.

darnuria:
## Related issue

Attempt to implement this suggestion <https://github.com/meilisearch/heed/issues/145#issuecomment-1384539350>

Alas the tracking `(device, inode)` only works for Unixes on windows it's more complicated. :/

A way could be to store https://docs.rs/same-file/latest/same_file but it don't expose something simple like `dev_t, ino_t`, but yet `Handle` implement `Hash` so maybe it may work (I am reviewing the implementation)!

## Caveat

`dev_t` and `ino_t` are unix concepts.


Kerollmops:
Thank you @darnuria,

It looks like the [`same_file::Handle` implements `Eq + Hash`](https://docs.rs/same-file/latest/same_file/struct.Handle.html#implementations), which makes it a good candidate for replacing the current key in the `OPENED_ENV` hashmap. What do you think?

heed/src/env.rs
```diff
@@ -180,22 +184,29 @@ impl EnvOpenOptions {
...
                 } else {
                     return Err(err.into());
                 }
             }
-            Ok(path) => path,
+            Ok(path) => {
+                let handle = Handle::from_path(&path)?;
+                (path, handle)
+            }
         };
```
darnuria, line 207: @Kerollmops I am wondering if we still need canonisation, for `Flag::NoSubDir` it's mandatory I think, but for the other possibilities?

heed/src/env.rs
```diff
@@ -277,9 +288,11 @@ impl EnvOpenOptions {
 }
 
 /// Returns a struct that allows to wait for the effective closing of an environment.
-pub fn env_closing_event<P: AsRef<Path>>(path: P) -> Option<EnvClosingEvent> {
+pub fn env_closing_event<P: AsRef<Path>>(path: P) -> Result<Option<EnvClosingEvent>> {
```
darnuria, line 291: It involve  public api changes here.

heed/src/env.rs
```diff
@@ -288,7 +301,7 @@ pub struct Env(Arc<EnvInner>);
 
 impl fmt::Debug for Env {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let EnvInner { env: _, dbi_open_mutex: _, path } = self.0.as_ref();
+        let EnvInner { env: _, dbi_open_mutex: _, path, .. } = self.0.as_ref();
         f.debug_struct("Env").field("path", &path.display()).finish_non_exhaustive()
     }
 }
```
darnuria, line 307: Shall I add to the debug the Handle?

heed/src/env.rs
```diff
@@ -853,6 +936,82 @@ mod tests {
...
+            .map_size(10 * 1024 * 1024) // 10MB
+            .open(&symlink_name)
+            .unwrap();
+
+        let _env = EnvOpenOptions::new()
+            .map_size(10 * 1024 * 1024) // 10MB
+            .open(&env_name)
+            .unwrap();
+    }
+
```
darnuria, line 1014: I don't possess a windows so I hope the CI or somebody would test it!

heed/src/env.rs
Kerollmops: Thank you, I added the `breaking` label and will create a release with the info.

heed/src/env.rs
Kerollmops: No, I don't see a reason to show it but could you please use the explicit syntax, here? i.e. `handle: _`.

Kerollmops (changes requested):

heed/src/env.rs
```diff
@@ -797,7 +811,7 @@ mod tests {
         eprintln!("env closed successfully");
 
         // Make sure we don't have a reference to the env
-        assert!(env_closing_event(&dir.path()).is_none());
+        assert!(env_closing_event(&dir.path()).unwrap().is_none());
```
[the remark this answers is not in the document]
darnuria: Did the `unwrap()` in tests since the api of `env_closing_env` changed to `Result<Option<EnvClosingEvent>>` Shall i avoid `unwrap` in tests?

No problem if it's the policy! :)

heed/src/env.rs
darnuria: We needed to go to the parent, to be frank I didn't thought this part preferred to open the discussion! :)

heed/src/env.rs
darnuria: Yeah good catch may avoid issue later!

heed/src/env.rs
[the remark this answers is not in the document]
Kerollmops: Ho! That's the tests sorry! I was in the subway, hard to review! All good 😄

heed/src/env.rs
Kerollmops: I am pretty sure that you can entirely remove the `canonicalize_path` function even when we use the `NoSubDir` flag. Could you create an issue to track this, please? We can prioritize this later (and specify that we must add strange tests about strange paths).

heed/src/env.rs
```diff
@@ -853,6 +922,82 @@ mod tests {
...
+        let symlink_name = dir_symlink.path().join("babar.mdb.link");
+
+        let mut envbuilder = EnvOpenOptions::new();
+        unsafe { envbuilder.flag(crate::Flag::NoSubDir) };
+        let _env = envbuilder
+            .map_size(10 * 1024 * 1024) // 10MB
+            .open(&env_name)
+            .unwrap();
+
+        std::os::unix::fs::symlink(&dir.path(), &symlink_name).unwrap();
```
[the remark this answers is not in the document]
darnuria: Fixed, messed-up the test in the first place now it's fixed would be glad to have  a second pair of eyes on this precise one.

heed/src/env.rs
```diff
@@ -29,34 +30,25 @@ use crate::{
...
+/// Trying to open a `None` marked environment returns an error to the user trying to open it.
+///
+/// [same_file::Handle] abstract the platform details of checking if the given paths points to
+///  the same file on the filesystem.
+///
+/// ## Safety
+///
+/// Mind that Handle currently open the file, so avoid writing through the fd held by [same_file::Handle],
+/// since the file will also be opened by LMDB.
+static OPENED_ENV: Lazy<RwLock<HashMap<Handle, EnvEntry>>> = Lazy::new(RwLock::default);
```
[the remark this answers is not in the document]
Kerollmops: Ok, so can we add a comment about it? I mean a really documentation comment in the lib.rs file. Under an Important/Remark header, please? I don't know how LMDB behaves towards that.

I am not sure that we really care [about this sentence](http://www.lmdb.tech/doc/) as we keep it open the whole env-lifetime right?
> Do not have open an LMDB database twice in the same process at the same time. Not even from a plain open() call - close()ing it breaks flock() advisory locking.

heed/src/env.rs
darnuria: https://github.com/meilisearch/heed/issues/180 don't hesitate to edit the issue and it's own scope.

heed/src/env.rs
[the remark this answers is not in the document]
darnuria: Yeah my worries comes from that line, and my first idea to just check `(dev_t, ino_t`)

For sure it may break the `flock()`ing (not the best system api..), and the multi-process features of lmdb is really handy.

I will try to checkout lmdb sources.

heed/src/env.rs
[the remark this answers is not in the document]
Kerollmops: Could you please for each test check that the path is the same? By making a comparison on the `Env::path`, you should see that the path is the one recycled from the previous env that is the same file.

BTW could you document, on `Env::path`, that the path output by `Env::path` is not necessarily the one given when opening the env but rather the one of the first env pointing to this file, please?

Kerollmops:
[I read somewhere](https://gitlab.com/chris-morgan/symlink#caution-symlinks-are-still-less-reliable-on-windows) that hard links on Windows require Administrator privileges...

heed/src/env.rs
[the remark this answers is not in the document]
darnuria: Another good catch will do it!

Kerollmops:
Do you know if we can run a certain pipeline in the CI as admin or Windows admin?

heed/src/env.rs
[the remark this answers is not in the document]
darnuria: > Could you please, for each test, check that the path is the same? By making a comparison, i.e. `assert_eq!` on the `Env::path`, you should see that the path is the one recycled from the previous env that points to the same file.

They are not recycled at the moment, think it's better to have the path local to the opening, what do you think?

but I agree to add `assert_eq!` to help maintaining the feature.

> BTW could you document, on `Env::path`, that the path output by `Env::path` is not necessarily the one given when opening the env but rather the one of the first env pointing to this file, please?

yeah sure!

heed/src/env.rs
```diff
@@ -830,6 +830,74 @@ mod tests {
...
+            .map_size(10 * 1024 * 1024) // 10MB
+            .open(&hardlink_name)
+            .unwrap();
+
+        let _env = envbuilder
+            .map_size(10 * 1024 * 1024) // 10MB
+            .open(&env_name)
+            .unwrap();
+    }
+
```
darnuria, line 857: This test is flawed, since we follow only `directory` and hardlink only apply to env it's just not-appliable.

Kerollmops: Can we have a symlink test with directories to replace this one? Obviously only running on UNIX.

heed/src/env.rs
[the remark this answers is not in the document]
Kerollmops: > ~They are not recycled at the moment, maybe it's better to have the path local to the opening?~ What do you think?~. Oh crap they are that PR is not ready yet (will point-out the faulty-code) Path defined here: https://github.com/meilisearch/heed/pull/179/files
> 
> but I agree to add `assert_eq!` to help maintaining the feature.

I don't understand this ☝

heed/src/env.rs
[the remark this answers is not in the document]
Kerollmops: So, what do we conclude? Can we still use the `same_file::Handle` struct? If not, what do we plan? Is it even possible to get the `dev_t` and `ino_t` without opening the file?

heed/src/env.rs
darnuria: Yeah already have one. :)
https://github.com/meilisearch/heed/pull/179/files#diff-ac577a0bca84e6fbce36b3a413107331b37c74809219009e740896600a1d97f7R861-R881

`symlink_dir` exist for windows two already did a test:
https://github.com/meilisearch/heed/pull/179/files#diff-ac577a0bca84e6fbce36b3a413107331b37c74809219009e740896600a1d97f7R982-R1002


May push the changes you suggested on test this morning.

heed/src/env.rs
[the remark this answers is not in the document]
darnuria: `stat` function family (except the `fstatat`) check metadata and return and don't maintain a file descriptor at least on unix.

It's the kind of function used by `metadata` in rust std, sadly posix never standardized a "device / FSID" so ino_t/dev_t is left to implementation detail.

heed/src/env.rs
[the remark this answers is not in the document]
darnuria: Added some `assert_eq!` to ensure we keep the path from the first opening. English not my mother tongue so sometimes it's messy sorry. ;)

darnuria:
> Do you know if we can run a certain pipeline in the CI as admin or Windows admin?

Don't know, also discovering that windows way of doing is really complicated. :clown_face:, like moving dir or getting file information.

heed/src/env.rs
```diff
@@ -830,6 +833,59 @@ mod tests {
...
+        fs::create_dir_all(&env_name).unwrap();
+
+        let env = EnvOpenOptions::new()
+            .map_size(10 * 1024 * 1024) // 10MB
+            .open(&env_name)
+            .unwrap();
+        assert_eq!(env_name, env.path());
+
+        let env_renamed = dir.path().join("serafina.mdb");
+        std::fs::rename(&env_name, &env_renamed).unwrap();
```
darnuria, line 880: for the windows part maybe the little UNIX api? https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/rename-wrename?view=msvc-170

PR #190, opened 2023-07-21
Replace the `&mut RwTxn` by `&RwTxn`

### LINKED ISSUE #189: We can't write the entries of one database into another
Currently, heed is too restrictive on the write transactions and do not permit certain basic operations like writing the content of one database into another as the example shows below.

[According to the documentation](https://github.com/LMDB/lmdb/blob/c7b3cc4df6dfe8f0772fb509bdc74777667caa43/libraries/liblmdb/mdb.c#L1396-L1397), write transactions were designed to support this kind of operations **on the same database** too.
 Which means that the following example could even work with the same database.

It would, therefore, be possible for heed replace the places where we use `&mut RwTxn` with, a less restrictive, `&RwTxn`.

```rust
for result in database1.iter(&wtxn)? {
    let (k, v) = result?;
    database2.put(&mut wtxn, k, v)?; // can't compile &mut and & of the same RwTxn used at the same time
}
```

Kerollmops:
The PR fixes #189 by removing all of the `&mut RwTxn` requirements by simple `&RoTxn`.

Doing so allows very cool use cases like:
 - being able to dump database entries into another one.
 - write into a database with values from the same database.

### DIFF
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -813,7 +813,7 @@ mod tests {
         let env = EnvOpenOptions::new()
             .map_size(10 * 1024 * 1024) // 10MB
             .max_dbs(30)
-            .open(&dir.path())
+            .open(dir.path())
             .unwrap();
 
         // Force a thread to keep the env for 1 second.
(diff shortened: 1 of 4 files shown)


Kerollmops:
Closed because it is unsafe to implement it like that. We must consider a more restrictive way to [let users copy the content from one database into another](https://github.com/meilisearch/heed/issues/189#issuecomment-1646822093).

PR #192, opened 2023-07-25
Make the `RoTxn: Send`

### LINKED ISSUE #191: Implement Send for RoTxn
I wanted to make the write transactions `Send` but LMDB's documentation says:
> A write [Transaction](https://lmdb.readthedocs.io/en/release/#lmdb.Transaction) may only be used from the thread it was created on.

However, I think that making `RoTxn: Sync` with the `sync-read-txn` is not safe, but I may be mistaken. But, for sure, `RoTxn` should implement `Send`, since it can be moved to other threads safely.

MarinPostma:
# Pull Request

## Related issue
Fixes #191

## What does this PR do?
We found out that implementing `Sync` for `RoTxn` is unsafe, it should instead only implement `Send`. This PR changes the `sync-read-txn` to `send-read-txn`, and replaces the implementation `Sync` for `RoTxn` with `Send`

## PR checklist

### DIFF
--- a/heed/Cargo.toml
+++ b/heed/Cargo.toml
@@ -36,7 +36,7 @@ url = "2.3.1"
 default = ["serde", "serde-bincode", "serde-json"]
 
 # The NO_TLS flag is automatically set on Env opening and
-# RoTxn implements the Sync trait. This allow the user to reference
+# RoTxn implements the Sync and Send traits. This allow the user to reference
 # a read-only transaction from multiple threads at the same time.
 sync-read-txn = []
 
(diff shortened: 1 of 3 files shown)


Kerollmops (changes requested):
Could you please create a new `send-read-txn` feature instead? I will remove the `sync-read-txn` in a future PR and try to do an ICE report about this feature.

Could you also add a link to [the LMDB documentation](https://github.com/LMDB/lmdb/blob/c7b3cc4df6dfe8f0772fb509bdc74777667caa43/libraries/liblmdb/lmdb.h#L610-L618) and explain that this new feature makes the `RoTxn: Send` and internally enables the `MDB_NOTLS` flag, please?

Kerollmops (approved):
Thank you very much @MarinPostma 🙏

I merge this PR as is but will probably:
 - Change the feature flag for it to be more general, e.g., `read-txn-no-tls`.
 - Add a lot more documentation on this specific feature.

PR #197, opened 2023-07-27
Remove the `PolyDatabase`

Kerollmops:
This PR removes the `PolyDatabase` type to simplify the library usage.

 - [x] One can always open a `Database` with newly introduced  `Unknown` types which don't implement the `BytesEncode/Decode` traits, forcing the user to `Database::remap_types`.
 - [x] I made the `Database::is_empty` method use the `Database::length` instead of allocating a cursor to see if there is at least one element. 
 - [x] I also introduced a new `DatabaseOpenOptions` struct to create or open a database akin to the `EnvOpenOptions` struct.
 - [x] I need to document the `DatabaseOpenOptions` struct.
 - [x] I moved the `LazyDecode` and `Lazy` types in the _heed-types_ crate.


Kerollmops:
Merge as part of #200.

PR #212, opened 2023-09-15
Support customized key compare function

### LINKED ISSUE #86: Support for custom key comparison function

xiaoyawei:
# Pull Request

## Related issue
Fixes #86 

## What does this PR do?

This PR continues the efforts in https://github.com/meilisearch/heed/pull/90 to allow users to customize key comparator. The implementation approaches stays mostly the same except the following parts

### Prefix iterator

`heed` implements the prefix iterator while `lmdb` does not actually have features like [prefix seek](https://github.com/facebook/rocksdb/wiki/Prefix-Seek) in rocksdb, which means all the logics that handles prefix are implemented in the rust wrapper layer.

After taking a closer look at the API of prefix iterator in `heed`, looks like currently `heed` has the following assumptions on the key order

1. For any `key` that starts with `prefix`, we have `prefix` < `key` unless `key` and `prefix` are identical.
2. If `prefix1` < `prefix2`, then for any `key1` that starts with `prefix1` and `key2` that starts with `prefix2`, we have `key1` < `key2`

These 2 assumptions together actually make the key order a lexicographic order, which I have a proof and can provide more details if needed. So on top of the `Comparator` trait, I also introduce a `LexicographicComparator` trait, and only when key order is lexicographic can users use the prefix iterator API.

### Database type

#86 mentions the attempt not to change the database type. In this PR, since I introduce the `LexicographicComparator`, I choose to add the comparator type as a generic type in `Database`'s definition, this brings the benefits that compile fails if a user is trying to use prefix iterator with a non-lexicographic comparator.

## Why do I need a custom key comparator?

Basically I am trying to implement a persistent key-ordered multimap on a bunch of storage engines. `LMDB` is faster than `rocksdb` with relatively small data volume so I'd love to add `LMDB` as an option for my multimap.

## PR checklist

@Kerollmops Since you are the author of the original PR and still an active committer in `heed`, I feel you must the be best person to take a look at this PR. I'd appreciate any feedback!

### DIFF
--- a/heed-traits/src/lib.rs
+++ b/heed-traits/src/lib.rs
@@ -3,6 +3,7 @@
 #![warn(missing_docs)]
 
 use std::borrow::Cow;
+use std::cmp::Ordering;
 use std::error::Error as StdError;
 
 /// A boxed `Send + Sync + 'static` error.
@@ -24,4 +25,55 @@ pub trait BytesDecode<'a> {
 
     /// Decode the given bytes as DItem
     fn bytes_decode(bytes: &'a [u8]) -> Result<Self::DItem, BoxedError>;
+}
+
+/// Define a custom key comparison function for a database.
+///
+/// The comparison function is called whenever it is necessary to compare a key specified
+/// by the application with a key currently stored in the database. If no comparison function
+/// is specified, and no special key flags were specified, the keys are compared lexically,
+/// with shorter keys collating before longer keys.
+pub trait Comparator {
+    /// Compares the raw bytes representation of two keys.
+    ///
+    /// # Safety
+    ///
+    /// This function must never crash, this is the reason why it takes raw bytes as parameter,
+    /// to let you define the recovery method you want in case of a decoding error.
+    fn compare(a: &[u8], b: &[u8]) -> Ordering;
+}
+
+/// Define a lexicographic comparator, which is a special case of [`Comparator`].
+///
+/// Types that implements [`LexicographicComparator`] will automatically have [`Comparator`]
+/// implemented as well, where the [`Comparator::compare`] is implemented per the definition
+/// of lexicographic ordering with [`LexicographicComparator::compare_elem`].
+///
+/// This trait is introduced to support prefix iterators, which implicit assumes that the
+/// underlying key comparator is lexicographic.
+pub trait LexicographicComparator: Comparator {
+    /// Compare a single byte; this function is used to implement [`Comparator::compare`]
+    /// by definition of lexicographic ordering.
+    ///
+    /// # Safety
+    ///
+    /// This function must never crash.
+    fn compare_elem(a: u8, b: u8) -> Ordering;
+
+    /// Update `bytes` to its immediate successor.
+    fn advance(bytes: &mut Vec<u8>);
+
+    /// Update `bytes` to its immediate predecessor.
+    fn retreat(bytes: &mut Vec<u8>);
+}
+
+impl<C: LexicographicComparator> Comparator for C {
+    fn compare(a: &[u8], b: &[u8]) -> Ordering {
+        for idx in 0..std::cmp::min(a.len(), b.len()) {
+            if a[idx] != b[idx] {
+                return a[idx].cmp(&b[idx]);
+            }
+        }
+        a.len().cmp(&b.len())
+    }
 }
--- a/heed-traits/src/lib.rs
+++ b/heed-traits/src/lib.rs
@@ -60,18 +60,28 @@ pub trait LexicographicComparator: Comparator {
     /// This function must never crash.
     fn compare_elem(a: u8, b: u8) -> Ordering;
 
-    /// Update `bytes` to its immediate successor.
-    fn advance(bytes: &mut Vec<u8>);
+    /// Advances the given `elem` to its immediate lexicographic successor, if possible.
+    /// Returns `None` if `elem` is already at its maximum value with respect to the
+    /// lexicographic order defined by this comparator.
+    fn advance(elem: u8) -> Option<u8>;
 
-    /// Update `bytes` to its immediate predecessor.
-    fn retreat(bytes: &mut Vec<u8>);
+    /// Moves the given `elem` to its immediate lexicographic predecessor, if possible.
+    /// Returns `None` if `elem` is already at its minimum value with respect to the
+    /// lexicographic order defined by this comparator.
+    fn retreat(elem: u8) -> Option<u8>;
+
+    /// Returns the maximum byte value per the comparator's lexicographic order.
+    fn max_elem() -> u8;
+
+    /// Returns the minimum byte value per the comparator's lexicographic order.
+    fn min_elem() -> u8;
 }
 
 impl<C: LexicographicComparator> Comparator for C {
     fn compare(a: &[u8], b: &[u8]) -> Ordering {
         for idx in 0..std::cmp::min(a.len(), b.len()) {
             if a[idx] != b[idx] {
-                return a[idx].cmp(&b[idx]);
+                return C::compare_elem(a[idx], b[idx]);
             }
         }
         a.len().cmp(&b.len())
--- a/heed-traits/src/lib.rs
+++ b/heed-traits/src/lib.rs
@@ -3,7 +3,7 @@
 #![warn(missing_docs)]
 
 use std::borrow::Cow;
-use std::cmp::Ordering;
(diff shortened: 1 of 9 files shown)


heed/src/database.rs
```diff
@@ -52,14 +54,14 @@ use crate::*;
...
+pub struct DatabaseOpenOptions<'e, KC, DC, C> {
     env: &'e Env,
-    types: marker::PhantomData<(KC, DC)>,
+    types: marker::PhantomData<(KC, DC, C)>,
     name: Option<String>,
     flags: AllDatabaseFlags,
 }
 
-impl<'e> DatabaseOpenOptions<'e, Unspecified, Unspecified> {
+impl<'e> DatabaseOpenOptions<'e, Unspecified, Unspecified, Unspecified> {
```
Kerollmops, line 64: We should probably use the `NoopComparator`, what do you think?

heed/src/iterator/range.rs
```diff
@@ -32,12 +31,13 @@ fn move_on_range_start<'txn>(
...
-            retreat_key(start);
-            result
-        }
+        Bound::Excluded(start) => match cursor.move_on_key_greater_than_or_equal_to(start)? {
+            Some((key, _)) if key == start => {
+                cursor.move_on_next(MoveOperation::NoDup)?;
+                cursor.move_on_prev(MoveOperation::NoDup)
+            }
+            result => Ok(result),
+        },
```
Kerollmops, line 40: It's probably better but are we sure about edge cases?

heed/src/env.rs
```diff
@@ -552,7 +599,12 @@ impl Env {
 
         // safety: The name cstring is cloned by LMDB, we can drop it after.
         //         If a read-only is used with the MDB_CREATE flag, LMDB will throw an error.
-        unsafe { mdb_result(ffi::mdb_dbi_open(raw_txn, name_ptr, flags, &mut dbi))? };
+        unsafe {
+            mdb_result(ffi::mdb_dbi_open(raw_txn, name_ptr, flags, &mut dbi))?;
+            if TypeId::of::<C>() != TypeId::of::<NoopComparator>() {
```
Kerollmops, line 604: Ok, that's smart! I was thinking about the way you were ignoring the `NoopComparator`.

heed/src/env.rs
```diff
@@ -333,6 +335,49 @@ impl Drop for EnvInner {
...
+        Ordering::Less => -1,
+        Ordering::Equal => 0,
+        Ordering::Greater => 1,
+    }
+}
+
+/// This is a special implementation of [`LexicographicComparator`] which means that no customized
+/// compare function is set to the underlying LMDB database instance. Its implementation of
+/// [`Comparator::compare`] will not be actually used.
+pub struct NoopComparator;
```
Kerollmops, line 356: I would like to change the name of this to something like `NoCustomComparator`, `IgnoreComparator`, `LexicograhicComparator` or `DefaultComparator` as it is in fact a type that is made to use the default LMDB lexicographic comparator one. Do you have better name ideas?

heed/src/env.rs
```diff
@@ -333,6 +335,49 @@ impl Drop for EnvInner {
...
+
+    fn retreat(bytes: &mut Vec<u8>) {
+        match bytes.last_mut() {
+            Some(&mut 0) => {
+                bytes.pop();
+            }
+            Some(last) => *last -= 1,
+            None => panic!("Vec is empty and must not be"),
+        }
+    }
```
Kerollmops, line 378: I thought we were not using these as the `NoopComparator` is not used but we actually use the `advance` and `retreat` methods when doing the prefix methods. Can we add a comment about this, please?

heed/src/database.rs
```diff
@@ -71,12 +73,23 @@ impl<'e> DatabaseOpenOptions<'e, Unspecified, Unspecified> {
...
+            env: self.env,
+            types: Default::default(),
+            name: self.name,
+            flags: self.flags,
+        }
+    }
+    /// Change the customized key compare function of the database.
+    ///
+    /// By default no customized compare function will be set when opening a database.
+    pub fn comparator<NC>(self) -> DatabaseOpenOptions<'e, KC, DC, NC> {
```
Kerollmops, line 92: Should we rename this method into `key_comparator`, what do you think?

heed/examples/custom_comparator.rs
```diff
@@ -0,0 +1,61 @@
+use std::cmp::Ordering;
+use std::error::Error;
+use std::path::Path;
+use std::{fs, str};
+
+use heed::EnvOpenOptions;
+use heed_traits::Comparator;
+use heed_types::{Str, Unit};
+
+struct StringAsIntCmp;
```
Kerollmops, line 10:
```suggestion
enum StringAsIntCmp {}
```

heed-traits/src/lib.rs
```diff
@@ -25,3 +26,54 @@ pub trait BytesDecode<'a> {
...
+}
+
+impl<C: LexicographicComparator> Comparator for C {
+    fn compare(a: &[u8], b: &[u8]) -> Ordering {
+        for idx in 0..std::cmp::min(a.len(), b.len()) {
+            if a[idx] != b[idx] {
+                return a[idx].cmp(&b[idx]);
+            }
+        }
+        a.len().cmp(&b.len())
```
Kerollmops, line 77: Shouldn't we use [the `[u8]::cmp`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.cmp) method here?
```suggestion
        a.cmp(b)
```

Kerollmops (changes requested):
Thank you very much for your changes, it seems great and working well. I just have some small change proposals.

heed/src/database.rs
xiaoyawei: Good call! Just updated the code by following your suggestion

heed/src/env.rs
xiaoyawei: Thanks!

heed/src/env.rs
xiaoyawei: I'll go with `DefaultComparator`; besides I add some more comprehensive comments to make sure most people won't get confused with it

heed/src/env.rs
xiaoyawei: Good call! I added it in the comments for `DefaultComparator`

heed/src/database.rs
xiaoyawei: Great idea, and code is updated!

heed-traits/src/lib.rs
xiaoyawei: Use `Ord::Cmp` instead here, which is indeed more readable!

xiaoyawei:
@Kerollmops, thank you for your insightful feedback!

I've pushed two additional commits to this PR:
- The first one (52c0a60) implements bug fixes that emerged from further examination of potential edge cases within this PR. New test cases have been added to confirm these fixes, which interestingly, reveal failures in the current `master` branch. Therefore, this commit not only enhances the PR but also rectifies existing issues in the main codebase. Details can be found within the test cases themselves.
- The second commit (711d7c4) incorporates your review comments and suggestions.

I welcome any additional feedback or recommendations you might have. Thank you!

heed/src/iterator/range.rs
Kerollmops: That's amazing! I mean, that you fixed some bugs! Thank you very much.

Kerollmops (approved):

Kerollmops:
Thank you very much for your PR and fixes on that one! I am very happy to merge this PR!
I will immediately release a v0.20.0-alpha.5 for you or others to try 🎉

PR #221, opened 2023-11-16
Introduce the `remap` method on the `Lazy` type

Kerollmops:
This PR introduces a new method to change the codec of a `LazyDecode`d type. The `Lazy` type returned can be set to decode into another type on demand and doesn't consume the original `Lazy` type; Therefore, it can be first decoded into something to change the way it is effectively decoded later on.


irevoire (approved):
nice,
bors merge

PR #241, opened 2024-02-16
Make the iterators `Send` when `read-txn-no-tls` is enabled

### LINKED ISSUE #213: Make `RoIter` also `Send` when `read-txn-no-tls` feature is enabled
With similar reasoning that `RoTxn` can be moved among threads when `MDB_NOTLS` flag is on, my understanding it that the corresponding `RoCursor` and `RoIter` (and other readonly iterators like range iterators) can be moved among threads as well, and hereby can implement `Send`.

By looking at the source code of `lmdb`, looks like [cursors is actually a field of a transaction](https://github.com/LMDB/lmdb/blob/mdb.master/libraries/liblmdb/mdb.c#L1335), and since different threads can access to the same db with the transaction pointer not simultaneously, they are also able to use a specific cursor in that transaction for data access.

I could be mistaken but would love to hear about opinions from `heed` maintainers. Thanks!

Kerollmops:
This PR closes #213 by making sure the iterators implement `Send` when the `read-txn-no-tls` feature is enabled.

### DIFF
--- a/heed/Cargo.toml
+++ b/heed/Cargo.toml
@@ -35,11 +35,11 @@ url = "2.3.1"
 default = ["serde", "serde-bincode", "serde-json"]
 serde = ["bitflags/serde", "dep:serde"]
 
-# The #MDB_NOTLS flag is automatically set on Env opening and
-# RoTxn implements the Send trait. This allows the user to move
-# a RoTxn between threads as read transactions will no more use
-# thread local storage and will tie reader locktable slots to
-# #MDB_txn objects instead of to threads.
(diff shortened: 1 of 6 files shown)


Kerollmops:
Hey @xiaoyawei 👋 Can you review this PR, please? It's small 😄

xiaoyawei:
@Kerollmops Thanks for putting this together and it all LGTM.

Looks like I don't have permission to approve a PR in this repository since I don't see a button to "approve". :)

xiaoyawei (approved):
LGTM

Issue #207, opened 2023-08-24
Opening an `Env` must be `unsafe`


Kerollmops:
[According to some comments](https://old.reddit.com/r/rust/comments/15z6mhv/release_v0200alpha4_of_heed_the_most_maintained/jxjjsai/), we should mark the opening of the LMDB environment as `unsafe`. I found out that [most people agree on that](https://users.rust-lang.org/t/how-unsafe-is-mmap/19635), and we could list the issues that opening an environment with memory-mapping could bring by using [the official documentation](http://www.lmdb.tech/doc/index.html).

zoo868e:
Hi @Kerollmops, I would like to try this. However, I can't compile the project. Here is the output of `cargo check`
```
   Compiling lmdb-master-sys v0.1.0 (/home/matt_jan/heed/lmdb-master-sys)
The following warnings were emitted during compilation:

warning: In file included from /usr/local/include/linux/stat.h:6,
warning:                  from /usr/include/x86_64-linux-gnu/bits/statx.h:31,
warning:                  from /usr/include/x86_64-linux-gnu/sys/stat.h:465,
warning:                  from /home/matt_jan/heed/lmdb-master-sys/lmdb/libraries/liblmdb/mdb.c:102:
warning: /usr/local/include/asm/stat.h:8:10: fatal error: uapi/asm/stat.h: No such file or directory
warning:     8 | #include <uapi/asm/stat.h>
warning:       |          ^~~~~~~~~~~~~~~~~
warning: compilation terminated.

error: failed to run custom build command for `lmdb-master-sys v0.1.0 (/home/matt_jan/heed/lmdb-master-sys)`

Caused by:
  process didn't exit successfully: `/home/matt_jan/heed/target/debug/build/lmdb-master-sys-aca218c1dc508df9/build-script-build` (exit status: 1)
  --- stdout
  TARGET = Some("x86_64-unknown-linux-gnu")
  OPT_LEVEL = Some("0")
  HOST = Some("x86_64-unknown-linux-gnu")
  cargo:rerun-if-env-changed=CC_x86_64-unknown-linux-gnu
  CC_x86_64-unknown-linux-gnu = None
  cargo:rerun-if-env-changed=CC_x86_64_unknown_linux_gnu
  CC_x86_64_unknown_linux_gnu = None
  cargo:rerun-if-env-changed=HOST_CC
  HOST_CC = None
  cargo:rerun-if-env-changed=CC
  CC = None
[...]

PR #245, opened 2024-03-01
Add `Env::set_flags()` and `FlagSetMode`

hinto-janai:
Implements https://github.com/meilisearch/heed/issues/202.

This PR:
1. Exposes `mdb_env_set_flags()` via FFI which gets used in...
2. `Env::set_flags()` which enables/disables flags depending on...
3. the `FlagSetMode` enum which is either `Enable` or `Disable`

PR #246, opened 2024-03-01
Set `EnvOpenOptions::open()` to `unsafe`

hinto-janai:
Closes https://github.com/meilisearch/heed/issues/207.

- `fn EnvOpenOptions::open()` -> `unsafe fn EnvOpenOptions::open()`
- Adds `# Safety` documentation
- Fixes all callers to use `unsafe`

`heed` can actually uphold some of the safety invariants, but for now they are documented so at least users are aware of them.

Unrelated change: the new `Env::set_flags` (https://github.com/meilisearch/heed/pull/245) doesn't necessarily gain safety from `&mut self`, so it is changed to `&self`.

### DIFF
--- a/heed/examples/all-types.rs
+++ b/heed/examples/all-types.rs
@@ -12,10 +12,12 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     fs::create_dir_all(&path)?;
 
-    let env = EnvOpenOptions::new()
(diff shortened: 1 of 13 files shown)


Kerollmops (changes requested):

Kerollmops (approved):
Thank you very much for the changes! Merging!

PR #251, opened 2024-04-28
replace all `FnMut` with `FnOnce` in `put_reserve_*`

nolanderc:
# Pull Request

## Related issue
N/A

## What does this PR do?
- Replaces `FnMut` with `FnOnce` in all `put_reserve_*` functions.
    - A value should only be written once, so `FnMut` is needlessly restrictive; with `FnOnce` we allow more functions to be used.

## PR checklist

### DIFF
--- a/heed/src/cursor.rs
+++ b/heed/src/cursor.rs
@@ -334,10 +334,10 @@ impl<'txn> RwCursor<'txn> {
         flags: PutFlags,
         key: &[u8],
         data_size: usize,
-        mut write_func: F,
+        write_func: F,
     ) -> Result<bool>
     where
-        F: FnMut(&mut ReservedSpace) -> io::Result<()>,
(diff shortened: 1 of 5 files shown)


Kerollmops (approved):
Thank you very much 👌

nolanderc:
> I am wondering if this is a breaking change 🤔 Because it is not constraining **more**....

It shouldn't be: any code which previously passed a FnMut closure will still compile as all FnMut closures also implement FnOnce.

Granted, this is a signature change, so if somebody passed the function pointer to another function `foo(Database::put_reserved)`, where `foo` had an FnMut bound on the `write_value` parameter, this would no longer compile. So technically this is breaking, but I can’t think of why you would do something like that…

Kerollmops:
Ok, so we can call that a soft-breaking change, and therefore, it's allowed 🤭 Thank you. It will be released in v0.20.1, then.

PR #256, opened 2024-04-29
make `ReservedSpace` more flexible

### LINKED ISSUE #254: Make `ReservedSpace` more flexible
The only way to use [`ReservedSpace`](https://docs.rs/heed/latest/heed/struct.ReservedSpace.html) currently is to use the `std::io::Write` implementation. However, this assumes that the entire value can be written in one go. This prevents certain patterns, such as having a header in the value which contains the a checksum of the remaining bytes.

There should be some mechanism to reference arbitrary bytes within the space. Some options (naming TBD):

1. Add a method to get all previously written bytes: `fn written_bytes_mut(&mut self) -> &mut [u8]`. This is safe as the bytes are known to be initialized.
2. Add a method to zero-initialize and get a mutable view of the entire space: `fn zero_bytes_mut(&mut self) -> &mut [u8]`. This would mark the entire space as written.
3. Add a method `fn as_bytes_mut_uninit(&mut self) -> &mut [MaybeUninit<u8>]`. Can be used to write arbitrary data in any order. Needs a complementary method to mark them as written: `unsafe fn assume_written(&mut self, count: usize)`, which is `unsafe` because caller must guarantee that all bytes `0..count` have been initialized.

nolanderc:
# Pull Request

## Related issue
Fixes #254 

## What does this PR do?
- Adds a few methods to `ReservedSpace` which grant access to the underlying bytes without having to go through the `Write` impl.
- Adds `impl std::io::Seek for ReservedSpace` which allows seeking within the reserved space and thus re-write parts of the space already written.

## PR checklist

### DIFF
--- a/heed/src/reserved_space.rs
+++ b/heed/src/reserved_space.rs
@@ -9,14 +9,16 @@ use crate::mdb::ffi;
 pub struct ReservedSpace {
     /// Total size of the space.
     size: usize,
-
     /// Pointer to the start of the reserved space.
     start_ptr: *mut u8,
-
     /// Number of bytes which have been written: all bytes in `0..written`.
     written: usize,
-
     /// Index of the byte which should be written next.
+    ///
+    /// # Safety
+    ///
+    /// To ensure there are no unwritten gaps in the buffer this should be kept in the range
+    /// `0..=written` at all times.
     write_head: usize,
 }
 
@@ -60,7 +62,7 @@ impl ReservedSpace {
     /// After calling this function, the entire space is considered to be filled and any
     /// further attempt to [`write`](std::io::Write::write) anything else will fail.
     pub fn fill_zeroes(&mut self) {
-        for i in self.written..self.size {
+        for i in self.write_head..self.size {
             unsafe { self.start_ptr.add(i).write(0) };
         }
         self.written = self.size;
@@ -121,7 +123,7 @@ impl io::Write for ReservedSpace {
 
 /// ## Note
 ///
-/// May only seek within the already within the previously written space.
+/// May only seek within the previously written space.
 /// Attempts to do otherwise will result in an error.
 impl io::Seek for ReservedSpace {
     fn seek(&mut self, pos: io::SeekFrom) -> io::Result<u64> {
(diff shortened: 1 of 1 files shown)


heed/src/reserved_space.rs
```diff
@@ -23,27 +37,129 @@ impl ReservedSpace {
...
+
     fn flush(&mut self) -> io::Result<()> {
         Ok(())
     }
 }
 
+/// ## Note
+///
+/// May only seek within the already within the previously written space.
+/// Attempts to do otherwise will result in an error.
```
Kerollmops, line 125:
```suggestion
/// May only seek within the previously written space.
/// Attempts to do otherwise will result in an error.
```

heed/src/reserved_space.rs
```diff
@@ -6,14 +7,27 @@ use crate::mdb::ffi;
...
     size: usize,
+
+    /// Pointer to the start of the reserved space.
     start_ptr: *mut u8,
+
+    /// Number of bytes which have been written: all bytes in `0..written`.
     written: usize,
+
+    /// Index of the byte which should be written next.
+    write_head: usize,
```
Kerollmops, line 20:
```suggestion
    /// Total size of the space.
    size: usize,
    /// Pointer to the start of the reserved space.
    start_ptr: *mut u8,
    /// Number of bytes which have been written: all bytes in `0..written`.
    written: usize,
    /// Index of the byte which should be written next.
    write_head: usize,
```

Kerollmops (commented):
Thank you very much for your work!

However, this `ReservedSpace` struct and API do not seem as optimized and performance optimal as I wanted. I [created a small reproduction of the issue](https://github.com/Kerollmops/heed-perfs-put-reserved) and an example of usage at Meilisearch. Would you mind taking some time to see if you can find the design issue in this API?

heed/src/reserved_space.rs
```diff
@@ -60,7 +62,7 @@ impl ReservedSpace {
     /// After calling this function, the entire space is considered to be filled and any
     /// further attempt to [`write`](std::io::Write::write) anything else will fail.
     pub fn fill_zeroes(&mut self) {
-        for i in self.written..self.size {
+        for i in self.write_head..self.size {
```
Kerollmops, line 65: Ho! Nice catch 👏

Kerollmops (approved):
LGTM 👏  Merging this one for v0.20.1. Thank you again!

PR #260, opened 2024-05-03
Improve performance of `ReservedSpace`

nolanderc:
# Pull Request

## Related issue
Addresses (some) concerns raised in a comment: https://github.com/meilisearch/heed/pull/256#pullrequestreview-2030571441. 

## What does this PR do?
- Marks most methods of `ReservedSpace` as `#[inline]` allowing them to be inlined across crate boundaries. Especially for the `Write`-impl where callers may perform small writes, the overhead of not inlining can be significant.
- See https://github.com/Kerollmops/heed-perfs-put-reserved/pull/1 for some benchmarks and details.

## PR checklist
  - **Where can I find these?**

### DIFF
--- a/heed/src/reserved_space.rs
+++ b/heed/src/reserved_space.rs
@@ -120,7 +120,7 @@ impl io::Write for ReservedSpace<'_> {
         unsafe {
             // SAFETY: we can always cast `T` -> `MaybeUninit<T>` as it's a transparent wrapper
             let buf_uninit = std::slice::from_raw_parts(buf.as_ptr().cast(), buf.len());
-            remaining.get_unchecked_mut(..buf.len()).copy_from_slice(buf_uninit);
+            remaining.as_mut_ptr().copy_from_nonoverlapping(buf_uninit.as_ptr(), buf.len());
         }
 
         self.write_head += buf.len();
--- a/heed/src/reserved_space.rs
+++ b/heed/src/reserved_space.rs
@@ -111,17 +111,17 @@ impl io::Write for ReservedSpace<'_> {
 
     #[inline]
     fn write_all(&mut self, buf: &[u8]) -> io::Result<()> {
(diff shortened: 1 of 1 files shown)


heed/src/reserved_space.rs
```diff
@@ -97,24 +103,33 @@ impl ReservedSpace<'_> {
...
+        }
+
+        unsafe {
+            // SAFETY: we can always cast `T` -> `MaybeUninit<T>` as it's a transparent wrapper
+            let buf_uninit = std::slice::from_raw_parts(buf.as_ptr().cast(), buf.len());
+            remaining.as_mut_ptr().copy_from_nonoverlapping(buf_uninit.as_ptr(), buf.len());
+        }
+
+        self.write_head += buf.len();
+        self.written = usize::max(self.written, self.write_head);
```
Kerollmops, line 127: Should we ignore this possible error? Shouldn't we either assert or return an I/O error, here?

Kerollmops (commented):
Thank you very much.

I [looked at the investigation](https://github.com/Kerollmops/heed-perfs-put-reserved/pull/1) and it seems the fastest implementation is the `put-reserved-uninit` method. What do you think?

Ho! And BTW, don't worry about the contribution guideline, this template is from the Meilisearch repo 😃

heed/src/reserved_space.rs
nolanderc: We already check for an error further up, this is just making sure that we correctly keep track of which bytes have been initialized (`written`). 

For example:
```rust
write(7);
seek(3);
write(2);
```
Should result in `written == 7`, because we moved the `write_head` back to `3` in `seek`, and then advanced it to `5`, but the bytes in `0..7` have still been initialized, so using `max` we ensure we don’t ”unmark” bytes as initialized.

But maybe I misunderstood the question?

PR #261, opened 2024-05-06
enable self referential rtxn

irevoire:
Enables rtxn that owns their env with static lifetimes.

That can be practical when your env is stored in the same struct that holds the `Database` and you don't want to create a self-referential struct


Kerollmops (approved):
I do like this PR. Thank you very much ❤️

PR #263, opened 2024-05-24
Add the "mdb_maxkeysize_0" feature which sets "-DMDB_MAXKEYSIZE=0" when compiling lmdb

tpunder:
This PR (which does not have a corresponding issue) adds a `mdb_maxkeysize_0` feature to `lmdb-master-sys` and `heed` which sets `-DMDB_MAXKEYSIZE=0` when compiling LMDB. This allows you to use keys that are larger than the default of 511 bytes long.

The feature is 100% opt-in.

I've added a conditional test to `lmdb-master-sys` and `heed` to check that keys larger than 511 can be successfully stored without returning an `MDB_BAD_VALSIZE` error.

Enabling this feature also allows you to use values larger than 511 in databases with the `MDB_DUPSORT` flag set (which was the problem I was running into).

Here is the documentation snippet from http://www.lmdb.tech/doc/group__internal.html:

> #define MDB_MAXKEYSIZE   (([MDB_DEVEL](http://www.lmdb.tech/doc/group__compat.html#ga103b045068a1d21bf2347a7342f8f486)) ? 0 : 511)
>
> The max size of a key we can write, or 0 for computed max.
>
> This macro should normally be left alone or set to 0. Note that a database with big keys or dupsort data cannot be reliably modified by a liblmdb which uses a smaller max. The default is 511 for backwards compat, or 0 when [MDB_DEVEL](http://www.lmdb.tech/doc/group__compat.html#ga103b045068a1d21bf2347a7342f8f486).
>
> Other values are allowed, for backwards compat. However: A value bigger than the computed max can break if you do not know what you are doing, and liblmdb <= 0.9.10 can break when modifying a DB with keys/dupsort data bigger than its max.
>
> Data items in an [MDB_DUPSORT](http://www.lmdb.tech/doc/group__mdb__dbi__open.html#gae0626566c2562e9007f5c8c9535bab1a) database are also limited to this size, since they're actually keys of a sub-DB. Keys and [MDB_DUPSORT](http://www.lmdb.tech/doc/group__mdb__dbi__open.html#gae0626566c2562e9007f5c8c9535bab1a) data items must fit on a node in a regular page.

### DIFF
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -901,6 +901,14 @@ impl Env {
         mdb_result(unsafe { ffi::mdb_env_set_mapsize(self.env_mut_ptr(), new_size) })
             .map_err(Into::into)
     }
+
+    /// Get the maximum size of keys and MDB_DUPSORT data we can write.
+    ///
+    /// Depends on the compile-time constant MDB_MAXKEYSIZE. Default 511
+    pub fn max_key_size(&self) -> usize {
+        let maxsize: i32 = unsafe { ffi::mdb_env_get_maxkeysize(self.env_mut_ptr()) };
+        maxsize as usize
+    }
 }
 
 /// Contains information about the environment.
@@ -1249,6 +1257,23 @@ mod tests {
             }
 
             env.prepare_for_closing().wait();
+        }
+    }
+
+    #[test]
+    fn max_key_size() {
+        let dir = tempfile::tempdir().unwrap();
+        let env = unsafe { EnvOpenOptions::new().open(dir.path().join(dir.path())).unwrap() };
+        let maxkeysize = env.max_key_size();
+
+        eprintln!("maxkeysize: {}", maxkeysize);
+
+        if cfg!(feature = "longer-keys") {
+            // Should be larger than the default of 511
+            assert!(maxkeysize > 511);
+        } else {
+            // Should be the default of 511
+            assert_eq!(maxkeysize, 511);
         }
     }
 }
--- a/heed/src/mdb/lmdb_error.rs
+++ b/heed/src/mdb/lmdb_error.rs
@@ -53,6 +53,14 @@ pub enum Error {
     /// Transaction cannot recover - it must be aborted.
     BadTxn,
     /// Unsupported size of key/DB name/data, or wrong DUP_FIXED size.
+    ///
+    /// Common causes of this error:
+    ///   - You tried to store a zero-length key
+    ///   - You tried to store a key longer than the max allowed key (511 bytes by default)
+    ///   - You are using [DUP_SORT](crate::DatabaseFlags::DUP_SORT) and trying to store a
+    ///     value longer than the max allowed key size (511 bytes by default)
+    ///
+    /// In the last two cases you can enable the `longer-keys` feature to increase the max allowed key size.
     BadValSize,
     /// The specified DBI was changed unexpectedly.
     BadDbi,
--- a/heed/Cargo.toml
+++ b/heed/Cargo.toml
@@ -90,6 +90,26 @@ mdb_idl_logn_14 = ["lmdb-master-sys/mdb_idl_logn_14"]
 mdb_idl_logn_15 = ["lmdb-master-sys/mdb_idl_logn_15"]
 mdb_idl_logn_16 = ["lmdb-master-sys/mdb_idl_logn_16"]
 
+# Setting this enables you to use keys longer than 511 bytes. The exact limit
+# is computed by LMDB at compile time. You can find the exact value by calling
+# Env::max_key_size(). This value varies by architecture.
+#
+# Example max key sizes:
+#   - Apple M1 (ARM64): 8126 bytes
+#   - Apple Intel (AMD64): 1982 bytes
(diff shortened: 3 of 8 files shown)


Kerollmops (changes requested):
Hey @tpunder 👋
What you've done is very cool! I am shocked that I was not aware of this feature before. Than you for your work.

However, could you rename the feature to something more user-firendly (newbie-friendly), please? Something in the vein of `read-txn-no-tls`. I am thinking about `longer-keys`, `infinite-key-length` or `unlimited-key-length`. Wat do you think?

I would also like if you could add a description with the advantages, as you describe them in this PR, to the feature of the heed crate (dupsort and stuff), and the limitations and dowsides if there are any (size increase to store length bytes, slowdowns).

BTW @AureliaDolo and @darnuria I don't know if you are aware of this but it fixes your problem with the key-size limits of LMDB when using the `DUPSORT` feature 🤩

heed/src/mdb/lmdb_error.rs
```diff
@@ -53,6 +53,13 @@ pub enum Error {
     /// Transaction cannot recover - it must be aborted.
     BadTxn,
     /// Unsupported size of key/DB name/data, or wrong DUP_FIXED size.
+    ///
+    /// Common causes of this error:
+    ///   - You tried to store a key longer than the max allowed key (511 bytes by default)
```
Kerollmops, line 58:
```suggestion
    ///   - You tried to store a zero-length key
```

heed/src/env.rs
```diff
@@ -901,6 +901,14 @@ impl Env {
         mdb_result(unsafe { ffi::mdb_env_set_mapsize(self.env_mut_ptr(), new_size) })
             .map_err(Into::into)
     }
+
+    /// Get the maximum size of keys and MDB_DUPSORT data we can write.
+    ///
+    /// Depends on the compile-time constant MDB_MAXKEYSIZE. Default 511
+    pub fn max_key_size(&self) -> usize {
+        let maxsize: i32 = unsafe{ ffi::mdb_env_get_maxkeysize(self.env_mut_ptr()) };
```
Kerollmops, line 909:
```suggestion
        let maxsize: i32 = unsafe { ffi::mdb_env_get_maxkeysize(self.env_mut_ptr()) };
```

Kerollmops (changes requested):
Thank you very much for the changes!

tpunder:
@Kerollmops I think `longer-keys` makes sense since you don't get unlimited key sizes. The actual max key length is architecture dependent (related to the page size). I've seen values of 8126 bytes on my M1 Macbook Pro and 1982 bytes on my Intel Macbook Pro.

I've updated the PR to include the rename. I've also hooked up the `mdb_env_get_maxkeysize` function (as `Env::max_key_size()`) so people have a way to check what the value is for them. I've also updated the docs a bit.

It's not clear to me that using `longer-keys` actually increases the overhead (e.g. length bytes) of any data stored on disk. It looks like a `short` is always used for the key length of a node: https://github.com/LMDB/lmdb/blob/88d0531d3ec3a592cdd63ca77647d31c568c24bc/libraries/liblmdb/mdb.c#L1124

I think the database/file format is the same regardless of the `MDB_MAXKEYSIZE`.

I think the only real downsize would be that the database files become potentially non-portable across architectures with different page sizes __if you are actually storing long keys__. For example: If I create a database on my M1 Macbook Pro (key length limit of 8126 bytes) with a key of size 2000 bytes and then try to open that file on my Intel Macbook pro (key length limit 1982) then I think I would have problems. However, if all of my keys are under 1982 bytes then I don't think there is any problem.

Kerollmops (changes requested):
Thank you very much for the investigation. Very appreciated 😇

Do you confirm that when using this feature, people will be able to store 2GiB data when opening a database with the `DUPSORT` feature? Or will the data length always be restricted to 8192/1900 bytes?

Will merge and do a release just after you fix the remaining issues.

heed/src/mdb/lmdb_error.rs
tpunder: I added the additional "You tried to store a zero-length key" bullet point. I'm not sure you meant to remove the "You tried to store a key longer than the max allowed key (511 bytes by default)" one.

heed/src/env.rs
tpunder: I fixed this spacing

heed/src/mdb/lmdb_error.rs
Kerollmops: Whoops, it's a mistake. Sorry, I forgot to copy/paste/edit, I just edited the line 😄

Kerollmops (changes requested):
Hey @tpunder 👋  I can merge only once the files are well formatted (check the CI). You can run a `cargo +nightly fmt` on the files you modified, especially _heed/src/mdb/lmdb_ffi.rs_ 😃

tpunder:
@Kerollmops I just ran `cargo +nightly fmt` to fix the formatting in `lmdb_ffi.rs`. Hopefully all is good now 😀

Kerollmops (approved):

Issue #257, opened 2024-04-29
Reduce allocations in `BytesEncode`


nolanderc:
The `BytesEncode` trait is currently defined as follows:

https://github.com/meilisearch/heed/blob/34063834019b41740b684a5aa3ebef8f872579f4/heed-traits/src/lib.rs#L20-L26

There are some issues with this definition:
- Unless the encoded type can be trivially transmuted to a slice of bytes (`str`, `[u8]`, `U32<NativeEndian>`, ...) we have to allocate a temporary buffer where the value is written.
- There is no way to get a temporary buffer without going through the global allocator.
- Allocating a temporary buffer incurs an extra copy.

For example, most architectures in use are Little-Endian (x86, Arm), however, for lexicographic ordering to correctly sort integers, we have to store them in Big-Endian. Since this incurs a byte-swap on Little-Endian architectures we cannot simply return a pointer to the integer itself, instead we have to allocate a temporary buffer where we store the byte-swapped result.

# Optimizations

In most cases we know a-priori the maximum size of the needed allocation:
- LMDB forces keys to be 511 bytes or less.
- Primitive types and arrays have a compile-time known size.

In some cases it may be preferable to run serialization in two phases:
- First we serialize to a "null" writer which counts the number of bytes needed for the value.
- With this size in hand, we can have LMDB reserve a memory area in the database which fits the serialized value (using `MDB_RESERVE`).
- We then serialize our value directly into the reserved space.

# A new trait

With the above in mind, we can make a few changes to the `BytesEncode` trait:

```rust
/// Set to `true` if the value can be trivially casted to `&[u8]` (e.g., `str`, `[u8]`, ...).
/// This is used as a hint by the library to avoid using intermediate buffers.
const ZERO_COPY: bool = false;

/// A hint to the database about the final size of the encoded value.
/// Can be used to reduce allocations and memory copies.
fn size_hint(item: &Self::EItem) -> Option<usize> {
    None
}

/// Encode the value into a pre-allocated buffer.
/// The buffer could be allocated on the stack if we are encoding keys (511 bytes),
/// or allocated on the heap (with an initial capacity equal to `size_hint`).
fn encode_writer<W: std::io::Write>(item: &Self::EItem, writer: &mut W) -> Result<()> {
    // the default impl forwards to the old method
    writer.write_all(Self::bytes_encode(item)?)?;
    Ok(())
}
```

[...]

PR #272, opened 2024-08-14
borrow primitive types where possible instead of allocating them on the heap in BytesEncode implementations

antonilol:
# Pull Request

## Related issue
somewhat related to #257

## What does this PR do?
Remove unnecessary allocations (heap allocations created with the `vec![...]` macro) in implementations of BytesEncode. Not all could be removed, without breaking the api, bytes need to be copied if they need to be flipped around (when using big endian integers on little endian machines).

### DIFF
--- a/heed-types/src/integer.rs
+++ b/heed-types/src/integer.rs
@@ -1,3 +1,4 @@
+use std::array;
 use std::borrow::Cow;
 use std::marker::PhantomData;
 use std::mem::size_of;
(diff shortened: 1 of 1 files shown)


Kerollmops:
Hey @antonilol 👋

Thank you for the great work but I am not ready to merge all of these `unsafe`. I would rather you propose a trait change by following the #257 instructions and my comment (to declare a `BytesEncode::zero_copy` method so that it's dynamic).

What do you think? We will then test that in Meilisearch to see the speed gain.

antonilol:
I am currently working on a new trait that is also (99%) backwards compatible with BytesEncode, this pr was more the 'quick fix' for what was possible with this current trait. The new one makes no heap allocations for primitive types, for both byte orders. Will make a (draft) pr with the current state of it!

PR #275, opened 2024-08-17
fix custom_key_cmp_wrapper being able to unwind to C code (ub)

antonilol:
`custom_key_cmp_wrapper` may not panic because it is called from C code that does not understand rust panics. If it does, that is undefined behavior.

`custom_key_cmp_wrapper` will now abort if the Comparator implementation panics. I also removed the safety comments on `compare` and `compare_elem` (the safety comments on the safe trait is what got me to discover this UB)

Review note:
```rust
match ord {
    Ordering::Less => -1,
    Ordering::Equal => 0,
    Ordering::Greater => 1,
}
```
is equivalent to `ord as i32`

### DIFF
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -373,6 +373,12 @@ impl Drop for EnvInner {
 
 /// A helper function that transforms the LMDB types into Rust types (`MDB_val` into slices)
 /// and vice versa, the Rust types into C types (`Ordering` into an integer).
+///
+/// # Safety
+///
+/// `a` and `b` should both properly aligned, valid for reads and should point to a valid
+/// [`MDB_val`][ffi::MDB_val]. An [`MDB_val`][ffi::MDB_val] (consists of a pointer and size) is
+/// valid when its pointer (`mv_data`) is valid for reads of `mv_size` bytes and is not null.
 unsafe extern "C" fn custom_key_cmp_wrapper<C: Comparator>(
     a: *const ffi::MDB_val,
     b: *const ffi::MDB_val,
--- a/heed/src/env.rs
+++ b/heed/src/env.rs
@@ -9,7 +9,9 @@ use std::os::unix::{
     ffi::OsStrExt,
     io::{AsRawFd, BorrowedFd, RawFd},
 };
+use std::panic::catch_unwind;
 use std::path::{Path, PathBuf};
+use std::process::abort;
 use std::sync::{Arc, RwLock};
 use std::time::Duration;
 #[cfg(windows)]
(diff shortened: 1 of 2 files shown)


heed/src/env.rs
```diff
@@ -371,16 +373,15 @@ impl Drop for EnvInner {
 
 /// A helper function that transforms the LMDB types into Rust types (`MDB_val` into slices)
 /// and vice versa, the Rust types into C types (`Ordering` into an integer).
-extern "C" fn custom_key_cmp_wrapper<C: Comparator>(
+unsafe extern "C" fn custom_key_cmp_wrapper<C: Comparator>(
```
Kerollmops, line 376: Why have you added unsafe to this function? Isn't it safe now?

heed/src/env.rs
```diff
@@ -371,16 +373,15 @@ impl Drop for EnvInner {
...
 ) -> i32 {
     let a = unsafe { ffi::from_val(*a) };
     let b = unsafe { ffi::from_val(*b) };
-    match C::compare(a, b) {
-        Ordering::Less => -1,
-        Ordering::Equal => 0,
-        Ordering::Greater => 1,
+    match catch_unwind(|| C::compare(a, b)) {
+        Ok(ord) => ord as i32,
+        Err(_) => abort(),
```
Kerollmops, line 384: I would rather prefer that we are explicit here. Can you keep the match, please?

Kerollmops (changes requested):
Thank you very much @antonilol to make the crate much more safe than it is already!

BTW what do you use heed for? 😊

heed/src/env.rs
antonilol: It still dereferences raw pointers. Should I add a `/// # Safety` doc comment with some explanation? It is a private function though.

antonilol:
> BTW what do you use heed for? 😊

I am currently experimenting with changing the database backend of [electrs](https://github.com/romanz/electrs), it currently uses RocksDB but that takes so long to compile.

heed/src/env.rs
Kerollmops: Ho! Yeah! You're right. It's a good idea to add a safety note.

antonilol: Added the safety note

Kerollmops (approved):
Thank you very much for the fixes! Merging 🙏

Kerollmops:
> I am currently experimenting with changing the database backend of [electrs](https://github.com/romanz/electrs), it currently uses RocksDB but that takes so long to compile.

Nice, @antonilol 👌 So, you switch not for the LMDB performances or functionalities but more to reduce compilation times?

antonilol:
> Nice, @antonilol 👌 So, you switch not for the LMDB performances or functionalities but more to reduce compilation times?

Mainly compile time, but the memory mapped nature of LMDB should also speed up runtime by a bit. Important functionality is about the same, talking about get, put, delete, and iterators.

PR #285, opened 2024-10-10
Make the `RoTxn` covariant and keep it !Sync

Kerollmops:
This PR makes the `RoTxn` and `RwTxn` covariant, and it is safe and keeps it `!Sync`.


dureuill (approved):
Was already covariant after all (my bad) but adding the doctest + using the NonNull is cleaner

PR #287, opened 2024-10-28
relax the HeedDecode bound on Lazy::remap

irevoire:
# Pull Request

## What does this PR do?
- Let user `remap` `Lazy` values that cannot be decoded _yet_


Kerollmops (approved):
That is very good! Thank you for that!

PR #289, opened 2024-11-27
add custom ordering support to range types

Boog900:
## What does this PR do?

This PR adds custom ordering support to the `iterator/range` types, by using the `Comparator` trait. Without this PR using any of the `range` functions on `Database` when a custom ordering is set will give unexpected results, like not giving all the keys in the range.

This PR also adds a `IntegerComparator` type, which represents the ordering when the `MDB_INTEGERKEY` flag is set. This is needed to allow the range functions to work when a table sets that flag.

### DIFF
--- a/heed/src/database.rs
+++ b/heed/src/database.rs
@@ -975,7 +975,7 @@ impl<KC, DC, C> Database<KC, DC, C> {
         }
     }
 
-    /// Return a lexicographically ordered iterator of all key-value pairs in this database.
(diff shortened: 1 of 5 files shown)


Kerollmops (approved):
Thank you very much for those changes!
I love seeing this kind of PR fixing hidden bugs like that.

As this is breaking, I'll probably merge this PR a little bit later when #278 lands.

PR #299, opened 2024-12-07
Introduce Checksumming

Kerollmops:
Takes most of the work from https://github.com/meilisearch/heed/pull/134/commits/b2cc01fc7a24d16bb2d9c707fe941af3b36f698b?show-viewed-files=true&file-filters%5B%5D=#diff-ac577a0bca84e6fbce36b3a413107331b37c74809219009e740896600a1d97f7

### DIFF
--- a/heed/Cargo.toml
+++ b/heed/Cargo.toml
@@ -1,29 +1,35 @@
 [package]
-name = "heed"
+name = "heed3"
 version = "0.21.0"
 authors = ["Kerollmops <renault.cle@gmail.com>"]
-description = "A fully typed LMDB (mdb.master) wrapper with minimum overhead"
+description = "A fully typed LMDB (mdb.master3) wrapper with minimum overhead with support for encryption"
 license = "MIT"
 repository = "https://github.com/Kerollmops/heed"
-keywords = ["lmdb", "database", "storage", "typed"]
+keywords = ["lmdb", "database", "storage", "typed", "encryption"]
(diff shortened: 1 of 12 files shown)


Kerollmops:
Hey @hyc 👋 I hope you're well.

While implementing checksumming support into my LMDB wrapper, I realized that I probably didn't understand some of the details about the checksumming feature.

I have a test that opens an LMDB environment from mdb.master3 and sets up checksumming by using `mdb_env_set_checksum` along with [a CRC algorithm](https://docs.rs/crc/latest/crc/constant.CRC_32_BZIP2.html). I write two entries, read them, commit, and close the environment. I modify the content on disk by also making sure not to change the length of the entries (even though it shouldn't be an issue) and reopen the database with the exact same settings. When I read the entries again, LMDB doesn't detect the corruption and returns the modified value.

So, I thought I needed to enable the encryption and checksumming features. I did that, and now LMDB is returning `BadTxn` when I try to read the entries just after the commit, not even after the on-disk on-purpose corruption, before I closed the environment.

I am thinking about a bug on LMDB's side and will create a reproducer if you confirm that this looks like one or if I forgot something. You can read [the test that is triggering the issue there](https://github.com/meilisearch/heed/blob/c560b51d2f7296cbc432652904902741eee1fabe/heed/src/envs/env_open_options.rs#L149-L233).

Have a nice day and end of the year, Howard ⛄

PR #303, opened 2025-01-07
Upgrade dependencies

Kerollmops:
This PR upgrades the compatible dependencies.

~After we merge this PR we need to publish a new version~ (We don't, it's only compatible upgrades).


irevoire (approved):
Nice

PR #304, opened 2025-01-08
Upgrade heed3 dependencies

Kerollmops:
Upgrade compatible and incompatible dependencies of heed3.

Note that I couldn't bump the `generic-array` dependency as the other ones use an old version.


irevoire (approved):
Perfecto :ok_hand:

PR #307, opened 2025-01-23
Introduce the `Env` and `RwTxn` `nested_read_txn` methods

Kerollmops:
This PR introduces two new methods in the heed and heed3 crates: the `Env::nested_read_txn` and `RwTxn::nested_read_txn`. These methods allow creating nested read-only transactions from an uncommitted one. It is powerful to parallelize the work on the content of a transaction, write the changes, and commit all the changes at once.

I wrote [a blog post on this subject](https://blog.kerollmops.com/patching-lmdb-how-we-made-meilisearch-s-vector-store-3x-faster) to explain in more detail how we use it on Meilisearch's side, and we discussed this in [ITS#10395](https://bugs.openldap.org/show_bug.cgi?id=10395) before finally merging the work in the upstream mdb.master3 branch. This is officially available to everyone using this experimental version of LMDB.

I decided to cherry-pick the changes on top of mdb.master, the official, feature-frozen version of LMDB that we use in Meilisearch, because we don't plan to use an experimental LMDB version and prefer to modify the LMDB version we already use internally just a bit. This PR has been merged, and the new versions of heed and heed3 are available on crates.io under the _v0.22.1_ tag.

```rust
// opening a write transaction

// [use the RwTxn to write into the env]

// opening multiple read-only transactions without committing beforehand
let rtxns = (0..1000).map(|_| env.nested_read_txn(&wtxn)).collect::<heed::Result<Vec<_>>>()?;

// you can also open a nested read transaction directly from the write transaction itself
let rtxns = (0..1000).map(|_| wtxn.nested_read_txn()).collect::<heed::Result<Vec<_>>>()?;

// [use the RoTxns and move them onto different threads]
```

### DIFF
--- a/heed/src/txn.rs
+++ b/heed/src/txn.rs
@@ -318,6 +318,73 @@ impl<'p> RwTxn<'p> {
         self.txn.inner.env.env_mut_ptr()
     }
 
+    /// Create a nested read transaction that is capable of reading uncommitted changes.
+    ///
+    /// The new transaction will be a nested transaction, with the transaction indicated by parent
+    /// as its parent. Transactions may be nested to any level.
+    ///
+    /// This is a custom LMDB fork feature that allows reading uncommitted changes.
+    /// It enables parallel processing of data across multiple threads through
+    /// concurrent read-only transactions. You can [read more in this PR](https://github.com/meilisearch/heed/pull/307).
+    ///
+    /// This method is equivalent to calling [`Env::nested_read_txn`] with this transaction as the parent.
+    ///
+    /// ```
+    /// use std::fs;
+    /// use std::path::Path;
+    /// use heed::{EnvOpenOptions, Database};
+    /// use heed::types::*;
+    ///
+    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// let dir = tempfile::tempdir()?;
+    /// let env = unsafe {
+    ///     EnvOpenOptions::new()
+    ///         .read_txn_without_tls()
+    ///         .map_size(2 * 1024 * 1024) // 2 MiB
+    ///         .open(dir.path())?
+    /// };
+    ///
+    /// // we will open the default unnamed database
+    /// let mut wtxn = env.write_txn()?;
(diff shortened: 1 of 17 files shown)


heed/src/txn.rs
```diff
@@ -84,6 +84,31 @@ impl<'e, T> RoTxn<'e, T> {
         Ok(RoTxn { txn: NonNull::new(txn), env: Cow::Owned(env), _tls_marker: PhantomData })
     }
 
+    #[cfg(not(master3))]
+    pub(crate) fn nested<'p>(
+        env: &'p Env<WithoutTls>,
+        parent: &'p RwTxn,
+    ) -> Result<RoTxn<'p, WithoutTls>> {
+        let mut txn: *mut ffi::MDB_txn = ptr::null_mut();
```
Kerollmops, line 92: Assert the env is the parent's one.

Kerollmops: Done in d99fc838865109f9bd2b60f599c458336dc07bb6

heed/src/txn.rs
```diff
@@ -100,6 +100,28 @@ impl<'e, T> RoTxn<'e, T> {
...
+        let parent_ptr: *mut ffi::MDB_txn = unsafe { parent.txn.inner.txn.unwrap().as_mut() };
+
+        unsafe {
+            // Note that we open a write transaction here and this is the (current)
+            // ugly way to trick LMDB and let me create multiple write txn.
+            mdb_result(ffi::mdb_txn_begin(env.env_mut_ptr().as_mut(), parent_ptr, 0, &mut txn))?
```
Kerollmops: I absolutely need to change this to include the RDONLY flag.

Kerollmops: Done 👌

Kerollmops:
We finally decided to merge my work upstream (in mdb.master3 still). The review was done in [ITS#10395](https://bugs.openldap.org/show_bug.cgi?id=10395). I decided to still patch mdb.master so that this feature is available to Meilisearch users without breaking the internal database. I will do a big squash of all the commits, plug the repository to the new upstream LMDB, and finally merge this work 👏

PR #309, opened 2025-02-03
Allow nested rtxn from wtxn for v0.20

Kerollmops:
This PR must not be merged as it is superseded by #307, which is much cleaner. This other one uses the runtime `WithTls` and `WithoutTls` types. The base branch is not even where I want to merge this.


Kerollmops:
Super-seeded by #307 and useless for Meilisearch as it is now using v0.22.

PR #315, opened 2025-03-04
Introduce `AnyTls`

### LINKED ISSUE #314: Ergonomics of `WithoutTls` and `WithTls`
Hello :wave:

I'm trying to integrate `heed` main on [a project of my company](https://github.com/meilisearch/arroy) 😎, but I find that the new `WithoutTls`/`WithTls` separation is a really big breaking change in the situation where we use a `&RwTxn` as a `RoTxn`.

The reason why is the following:

1. `RoTxn` defaults to `RoTxn<WithTls>`, so any function accepting a `&RoTxn` now actually accepts a `&RoTxn<WithTls>`
2. But the `RwTxn` derefs to `RoTxn<WithoutTls>`.

So, one has to go and annotate all functions that accept a `&RoTxn` to specify the `<WithoutTls>` part, even though it has no incidence on the code, as `RoTxn` is `!Sync` regardless of `WithoutTls`/`WithTls`.

My company has this [bigger project](https://github.com/meilisearch/meilisearch), and I'd find sad to update the 257 functions that accept a `&RoTxn` there.

---

Suggested resolution:

1. Introduce a third unit struct `AnyTls`.
2. Make `RoTxn` default to `RoTxn<AnyTls>`.
3. There is no actual way to build a `RoTxn<AnyTls>`. As a protection, `RoTxn<AnyTls>` is also `!Send` (but this should never matter).
4. But a `RoTxn<T>` always derefs to `RoTxn<AnyTls>`
5. And so does a `RwTxn`.

That way, any function accepting a `&RoTxn` now actually accepts a `&RoTxn<AnyTls>`, and no change is necessary for code that derefs a `&RwTxn` to a `&RoTxn`.

Thank you very much for making heed 😊, happy to send a PR your way

dureuill:
# Pull Request

## Related issue
Fixes #314 

## What does this PR do?
- Introduces `AnyTls`
- `RoTxn<T>` decay to `RoTxn<AnyTls>`
- To allow for transmutes, refactor `Env` and `RoTxn` to separate the phantom `T` (`WithTls`/`WithoutTls`/`AnyTls`) and the data + makes them repr transparent.
- Remove `T` from cursors: these types can never be `Send` because they morally have a reference to a `RoTxn`, which is not `Sync`.
- Removed the unsafe `Send` implementations of all the iterators.

### DIFF
--- a/heed/src/iterator/range.rs
+++ b/heed/src/iterator/range.rs
@@ -199,8 +199,6 @@ impl<KC, DC, C, IM> fmt::Debug for RoRange<'_, KC, DC, C, IM> {
     }
 }
 
-unsafe impl<KC, DC, C, IM> Send for RoRange<'_, KC, DC, C, IM> {}
-
 /// A read-write range iterator structure.
 pub struct RwRange<'txn, KC, DC, C = DefaultComparator, IM = MoveThroughDuplicateValues> {
     cursor: RwCursor<'txn>,
@@ -636,8 +634,6 @@ impl<KC, DC, C, IM> fmt::Debug for RoRevRange<'_, KC, DC, C, IM> {
         f.debug_struct("RoRevRange").finish()
     }
 }
-
-unsafe impl<KC, DC, C, IM> Send for RoRevRange<'_, KC, DC, C, IM> {}
 
 /// A reverse read-write range iterator structure.
 pub struct RwRevRange<'txn, KC, DC, C = DefaultComparator, IM = MoveThroughDuplicateValues> {
--- a/heed/src/iterator/range.rs
+++ b/heed/src/iterator/range.rs
@@ -8,8 +8,8 @@ use crate::cursor::MoveOperation;
 use crate::iteration_method::{IterationMethod, MoveBetweenKeys, MoveThroughDuplicateValues};
 use crate::*;
 
-fn move_on_range_end<'txn, T>(
-    cursor: &mut RoCursor<'txn, T>,
+fn move_on_range_end<'txn>(
+    cursor: &mut RoCursor<'txn>,
     end_bound: &Bound<Vec<u8>>,
 ) -> Result<Option<(&'txn [u8], &'txn [u8])>> {
     match end_bound {
@@ -25,8 +25,8 @@ fn move_on_range_end<'txn, T>(
     }
 }
 
-fn move_on_range_start<'txn, T>(
-    cursor: &mut RoCursor<'txn, T>,
+fn move_on_range_start<'txn>(
+    cursor: &mut RoCursor<'txn>,
     start_bound: &mut Bound<Vec<u8>>,
 ) -> Result<Option<(&'txn [u8], &'txn [u8])>> {
     match start_bound {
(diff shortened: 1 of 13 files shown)


heed/src/iterator/range.rs
```diff
@@ -199,7 +199,7 @@ impl<KC, DC, C, IM> fmt::Debug for RoRange<'_, KC, DC, C, IM> {
     }
 }
 
-unsafe impl<KC, DC, C, IM> Send for RoRange<'_, WithoutTls, KC, DC, C, IM> {}
+unsafe impl<KC, DC, C, IM> Send for RoRange<'_, KC, DC, C, IM> {}
```
Kerollmops, line 202: Hum... Do we want to keep this impl and the others that are similar?

heed/src/txn.rs
```diff
@@ -224,46 +270,41 @@ impl<'p> RwTxn<'p> {
...
         };
 
-        let env_without_tls = unsafe { env.as_without_tls() };
-
         Ok(RwTxn {
             txn: RoTxn {
-                txn: NonNull::new(txn),
-                env: Cow::Borrowed(env_without_tls),
+                inner: RoTxnInner { txn: NonNull::new(txn), env: Cow::Borrowed(&env.inner) },
+
```
Kerollmops, line 276:
```suggestion
```

heed/src/txn.rs
```diff
@@ -167,6 +209,10 @@ impl TlsUsage for WithoutTls {
     const ENABLED: bool = false;
 }
 
+impl TlsUsage for AnyTls {
+    const ENABLED: bool = false;
```
Kerollmops, line 213:
```suggestion
    // Users cannot open environments with AnyTls; therefore, this will never be read.
    const ENABLED: bool = false;
```

heed/src/txn.rs
```diff
@@ -109,15 +125,35 @@ impl<'e, T> RoTxn<'e, T> {
...
+        // SAFETY: OK because repr(transparent) means RoTxn<T> always has the same layout
+        // as RoTxnInner.
+        unsafe { std::mem::transmute(self) }
+    }
+}
+
 impl<T> Drop for RoTxn<'_, T> {
     fn drop(&mut self) {
-        if let Some(mut txn) = self.txn.take() {
+        if let Some(mut txn) = self.inner.txn.take() {
```
Kerollmops, line 156: Out of scope but should I really wrap this `NonNull` into an `Option` when I am sure that this struct cannot be dropped two times, it's impossible?

Kerollmops (dismissed):

heed/src/iterator/range.rs
Kerollmops: Moving Iterators is unsafe if the transaction uses TLS (`WithTls`). As we can no longer ensure that because we don't have the information in the type anymore, I wonder if we should reintroduce the information (and therefore a correct impl of this trait) or simply remove it (breaking).

Kerollmops: Ok, [we will have to reintroduce this `Send` impl](https://github.com/meilisearch/heed/issues/213) only when the underlying `RoTxn` is `Send` (`WithoutTls`).

Kerollmops: I reintroduced the possibility to move iterators between threads when `RoTxn` are using `WithoutTls`.

Kerollmops (approved):
Looks good to me!

PR #326, opened 2025-03-11
Introduce the `Env::copy_to_path` method

Kerollmops:
This PR introduces the `Env::copy_to_path` method that copies an env into a file created at the given path.

### DIFF
--- a/heed/src/envs/env.rs
+++ b/heed/src/envs/env.rs
@@ -1,6 +1,7 @@
 use std::any::TypeId;
 use std::ffi::CString;
-use std::fs::File;
+use std::fs::{self, File};
+use std::io::Seek;
 use std::marker::PhantomData;
 use std::path::{Path, PathBuf};
 use std::ptr::{self, NonNull};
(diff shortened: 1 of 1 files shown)


heed/src/envs/env.rs
```diff
@@ -441,6 +442,68 @@ impl<T> Env<T> {
         unsafe { self.copy_to_fd(fd, option) }
     }
 
+    /// Copy an LMDB environment to a file created at the given path, with options.
+    ///
+    /// This function may be used to make a backup of an existing environment.
+    /// No lockfile is created, since it gets recreated at need.
```
irevoire, line 448: What happens if there was already a lockfile? Shouldn’t we try to delete it?

irevoire (approved):
Small question about the lock file, but otherwise it’s all good both code-wise and comment-wise 👍

heed/src/envs/env.rs
Kerollmops: The lock file will not have any impact, we can just ignore it.

PR #327, opened 2025-03-11
Make the `Env<WithTls>` the default

Kerollmops:


dureuill:
Hello, what is the motivation here? I find that having to explicitly indicate which variant of env we want is good and rather unobstrusive

Kerollmops:
Hey 👋 It removes one breaking change and is similar to the change we did for the `RoTxn<T = WithTls>`.

dureuill (approved):
we'll have to be careful when integrating in Meilisearch to define all `Env` as `Env<WithoutTls>`. Otherwise, approved

irevoire:
What happens if we don’t? We should get an error when we try to share a RoTxn through multiple threads, right? It won’t silently crash?

Kerollmops:
Ahah, I already had to make them `Env<WithoutTls>` anyway and there are no actual issue if we forgot and require a `RoTxn<WithoutTls>`, it will simply not compile.

PR #330, opened 2025-03-19
Add function to remove a database

### LINKED ISSUE #214: provide fn to delete a database: mdb_drop(1)
You can clear a database but not remove it. This isn't high priority, but I like things tidy.

A-Manning:
# Pull Request

## Related issue
Fixes #214

## PR checklist

### DIFF
--- a/heed/src/databases/database.rs
+++ b/heed/src/databases/database.rs
@@ -2646,6 +2646,82 @@ impl<KC, DC, C, CDUP> Database<KC, DC, C, CDUP> {
         }
     }
 
+    /// Removes this database entirely.
+    ///
+    /// # Safety
+    ///
+    /// Ensure that no other copies of the database exist before calling, as
+    /// they will become invalid.
+    /// Do not remove a database if an existing transaction has modified it.
+    /// Doing so can cause database corruption or other errors.
+    ///
+    /// ```
+    /// # use std::fs;
+    /// # use std::path::Path;
+    /// # use heed::EnvOpenOptions;
+    /// use heed::Database;
+    /// use heed::types::*;
+    /// use heed::byteorder::BigEndian;
+    ///
+    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// # let dir = tempfile::tempdir()?;
+    /// # let env = unsafe { EnvOpenOptions::new()
+    /// #     .map_size(10 * 1024 * 1024) // 10MB
+    /// #     .max_dbs(3000)
+    /// #     .open(dir.path())?
+    /// # };
+    /// /// List databases in an env
+    #[cfg_attr(not(master3), doc = concat!(
+    "fn list_dbs(env: &heed::Env, rotxn: &heed::RoTxn<'_>) -> heed::Result<Vec<String>> {\n",
+    "    let names_db: Database<Str, DecodeIgnore> =",
+    ))]
+    #[cfg_attr(master3, doc = concat!(
+    "fn list_dbs(\n",
+    "    env: &heed::Env,\n",
+    "    rotxn: &heed::RoTxn<'_>,\n",
+    ") -> Result<Vec<String>, Box<dyn std::error::Error>> {\n",
+    "    // mdb-master3 uses null-terminated C strings as DB names\n",
+    "    let names_db: Database<Bytes, DecodeIgnore> =",
+    ))]
+    ///         env.open_database(&rotxn, None)?
+    ///            .expect("the unnamed database always exists");
+    ///     let mut names = Vec::new();
+    ///     for item in names_db.iter(&rotxn)? {
+    ///         let (name, ()) = item?;
+    #[cfg_attr(master3, doc = concat!(
+    "        let name = std::ffi::CStr::from_bytes_with_nul(name)?.to_str()?;",
+    ))]
+    ///         names.push(name.to_owned());
+    ///     }
+    ///     Ok(names)
+    /// }
+    ///
+    /// type BEI32 = I32<BigEndian>;
+    ///
+    /// let mut rwtxn = env.write_txn()?;
+    /// let db: Database<BEI32, Str> = env.create_database(&mut rwtxn, Some("iter-i32"))?;
+    /// rwtxn.commit()?;
+    ///
+    /// let rotxn = env.read_txn()?;
+    /// let db_names = list_dbs(&env, &rotxn)?;
+    /// assert_eq!(db_names, vec!["iter-i32".to_owned()]);
+    /// drop(rotxn);
+    ///
+    /// let mut rwtxn = env.write_txn()?;
+    /// unsafe { db.remove(&mut rwtxn)? };
+    /// let db_names = list_dbs(&env, &rwtxn)?;
+    /// assert!(db_names.is_empty());
+    /// rwtxn.commit()?;
+    /// # Ok(()) }
+    /// ```
+    pub unsafe fn remove(self, rwtxn: &mut RwTxn) -> Result<()> {
+        assert_eq_env_db_txn!(self, rwtxn);
+
+        unsafe {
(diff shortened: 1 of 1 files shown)


heed/src/databases/database.rs
```diff
@@ -2646,6 +2646,74 @@ impl<KC, DC, C, CDUP> Database<KC, DC, C, CDUP> {
         }
     }
 
+    /// Removes this database entirely.
+    /// The database is removed once the txn is committed.
```
Kerollmops, line 2650: That's better for the doc. The first sentence will be used when a short text is required.

```suggestion
    /// Removes this database entirely.
    ///
    /// The database is removed once the txn is committed.
```

heed/src/databases/database.rs
```diff
@@ -2646,6 +2646,74 @@ impl<KC, DC, C, CDUP> Database<KC, DC, C, CDUP> {
...
+    ///
+    /// # Safety
+    ///
+    /// Ensure that no other copies of the database exist before calling, as
+    /// they will become invalid.
+    /// Do not remove a database if an existing transaction has modified it.
+    /// Doing so can cause database corruption or other errors.
```
Kerollmops, line 2657: Where did you read that? Isn't the first sentence enough? Do you mean that removing a database is not thread safe?

heed/src/databases/database.rs
```diff
@@ -2646,6 +2646,74 @@ impl<KC, DC, C, CDUP> Database<KC, DC, C, CDUP> {
...
+    /// let mut rwtxn = env.write_txn()?;
+    /// unsafe { db.remove(&mut rwtxn)? };
+    ///
+    /// // DB still exists until the txn is committed
+    /// let db_names = list_dbs(&env)?;
+    /// assert_eq!(db_names, vec!["iter-i32".to_owned()]);
+    ///
+    /// rwtxn.commit()?;
+    /// let db_names = list_dbs(&env)?;
+    /// assert!(db_names.is_empty());
```
Kerollmops, line 2706: Would you mind changing this example a bit so that the read txn is created in the main function and not inside, please?

Kerollmops (changes requested):
Looks good overall, thank you 👌

heed/src/databases/database.rs
A-Manning: Which read txn are you referring to?

Kerollmops: Sorry, the one you open **inside** of `list_dbs`.

heed/src/databases/database.rs
A-Manning: See https://github.com/LMDB/lmdb/blob/b8e54b4c31378932b69f1298972de54a565185b1/libraries/liblmdb/lmdb.h#L1200-L1205

```
This call is not mutex protected. Handles should only be closed by
a single thread, and only if no other threads are going to reference
the database handle or one of its cursors any further. Do not close
a handle if an existing transaction has modified its database.
Doing so can cause misbehavior from database corruption to errors
like MDB_BAD_VALSIZE (since the DB name is gone).
```

IIUC, removing a DB should be thread-safe provided that no other existing txns have modified the DB, and no other copies of the DB hande exist.

heed/src/databases/database.rs
A-Manning: Done!

heed/src/databases/database.rs
Kerollmops: Ok, so let's make this function `unsafe` them, unfortunately.

heed/src/databases/database.rs
A-Manning: Done! This caused me to catch an error in my docs - the DB is deleted immediately in the rwtxn, so my comment about DB delete taking effect only after committing a rwtxn was incorrect. I have removed the comment.

heed/src/databases/database.rs
Kerollmops: Ahaha my bad. I'm on multiple fields right now 🤦

Kerollmops:
I don't know if this `\0` is expected and only visible on the heed3 part but you should probably read [this part of the README](https://github.com/meilisearch/heed?tab=readme-ov-file#working-with-two-crates-heed-and-heed3) to know how to deal with heed3.

A-Manning:
>I don't know if this \0 is expected and only visible on the heed3 part but you should probably read [this part of the README](https://github.com/meilisearch/heed?tab=readme-ov-file#working-with-two-crates-heed-and-heed3) to know how to deal with heed3.

It looks like DB names are stored differently in `mdb.master3`. From the [`lmdb_master3_sys` docs](https://docs.rs/lmdb-master3-sys/latest/lmdb_master3_sys/index.html):

> Note: Names are C strings and stored with their NUL terminator included. In LMDB 0.9 the NUL terminator was omitted.

It looks like the example needs to work slightly differently for `heed3` (as does the cookbook example for listing named databases). I'm not sure how that is supposed to happen in this codebase, are there any examples for writing different code for heed vs heed3?

Kerollmops:
Nice, I didn't know that! Good catch!
You can look at [this specific part of the code](https://github.com/meilisearch/heed/blob/31cf5dd0149ad786232e76d18ed44dfe153cc557/heed/src/envs/env_open_options.rs#L403). Basically if the `master3` cfg is enabled you are using heed3.

A-Manning:
I've added some `#[cfg_attr(..)]`s to generate a slightly different example for `heed3`. Tests are passing locally for both `heed` and `heed3`, can you run the CI again?

heed/src/databases/database.rs
```diff
@@ -2646,6 +2646,82 @@ impl<KC, DC, C, CDUP> Database<KC, DC, C, CDUP> {
...
+    "    let names_db: Database<Bytes, DecodeIgnore> =",
+    ))]
+    ///         env.open_database(&rotxn, None)?
+    ///            .expect("the unnamed database always exists");
+    ///     let mut names = Vec::new();
+    ///     for item in names_db.iter(&rotxn)? {
+    ///         let (name, ()) = item?;
+    #[cfg_attr(master3, doc = concat!(
+    "        let name = std::ffi::CStr::from_bytes_with_nul(name)?.to_str()?;",
+    ))]
```
Kerollmops, line 2693: Wow! Hard to believe it was that complex! Thank you for the help!

heed/src/databases/database.rs
```diff
@@ -2646,6 +2646,82 @@ impl<KC, DC, C, CDUP> Database<KC, DC, C, CDUP> {
...
+    ///
+    /// let mut rwtxn = env.write_txn()?;
+    /// unsafe { db.remove(&mut rwtxn)? };
+    /// let db_names = list_dbs(&env, &rwtxn)?;
+    /// assert!(db_names.is_empty());
+    /// rwtxn.commit()?;
+    /// # Ok(()) }
+    /// ```
+    pub unsafe fn remove(self, rwtxn: &mut RwTxn) -> Result<()> {
+        assert_eq_env_db_txn!(self, rwtxn);
```
Kerollmops, line 2718:
```suggestion
    pub unsafe fn remove(self, wtxn: &mut RwTxn) -> Result<()> {
        assert_eq_env_db_txn!(self, wtxn);
```

heed/src/databases/database.rs
```diff
@@ -2646,6 +2646,82 @@ impl<KC, DC, C, CDUP> Database<KC, DC, C, CDUP> {
...
+
+        unsafe {
+            mdb_result(ffi::mdb_drop(rwtxn.txn.txn_ptr().as_mut(), self.dbi, 1)).map_err(Into::into)
```
Kerollmops, line 2721:
```suggestion
    pub unsafe fn remove(self, wtxn: &mut RwTxn) -> Result<()> {
        assert_eq_env_db_txn!(self, wtxn);

        unsafe {
            mdb_result(ffi::mdb_drop(wtxn.txn.txn_ptr().as_mut(), self.dbi, 1)).map_err(Into::into)
```

Kerollmops (approved):
Let's merge! It looks good to me 😃 Thank you!

Issue #336, opened 2025-07-08
Heed3 : reading encrypted data in parallel threads leads to corruption


emuellen:
Hello, I recently upgraded to 0.22 and now I wanted to use the encrypted environment feature. Unfortunately, after just replacing all necessary signatures and changing the transactions to &mut, I get a lot of segmentation faults when running my tests. I already tried cloning the values I get the from lmdb to a vector as I was afraid that the internal buffer would be too small, but nothing helped. Do you know if running multiple threads with read transactions in an encrypted environment is supposed to work? Thanks a lot in advance for your response!

Here is an example that I use for my tests:

```
use std::{error::Error, sync::Arc, time::Instant};

use argon2::Argon2;

use chacha20poly1305::{ChaCha20Poly1305, Key};
use heed3::{
    types::{Bytes, Str},
    EnvOpenOptions,
};

use std::fs;

fn main() -> Result<(), Box<dyn Error + Send + Sync>> {
    let path_non_enc = "/tmp/lmdb_test_db_non_enc";
    let path = "/tmp/lmdb_test_db_enc";
    fs::create_dir_all(path_non_enc).expect("Failed to create path");
    fs::create_dir_all(path).expect("Failed to create path");

    let password = "thisisthepasswordasdfdasfasdf";
    let salt = "thisisthesalt123456789asdfasdffds";
    let mut key = Key::default();
    Argon2::default().hash_password_into(password.as_bytes(), salt.as_bytes(), &mut key)?;

    let start = Instant::now();

    //let env = unsafe { EnvOpenOptions::new().max_readers(20).map_size(1 << 32).max_dbs(5).open(path_non_enc)? };
    let env = unsafe { EnvOpenOptions::new().max_readers(20).map_size(1 << 32).max_dbs(5).open_encrypted::<ChaCha20Poly1305, _>(key, path)? };
    let env = Arc::new(env);

    let mut txn = env.write_txn()?;
    let db = env.create_database::<Str, Bytes>(&mut txn, Some("test_db"))?;
    for key in 0..1000 {
        db.put(&mut txn, &format!("{key}"), "toto".as_bytes()).unwrap();
    }
    txn.commit()?;

    let mut threads = Vec::new();
    for _ in 0..std::thread::available_parallelism().unwrap().get() {
        let env = env.clone();
        let thread = std::thread::spawn(move || {
            let mut txn = env.read_txn().unwrap();
[...]

PR #338, opened 2025-07-26
Fix encrypted heed3 parallel reads

Kerollmops:
This PR exposes the bug found in LMDB when trying to read an encrypted database from multiple threads at the same time #336. When running the example in Rust debug I get the following error message which tends to say that the Rust std slice is not happy about the pointer and/or arroy length it receives [in this function](https://github.com/meilisearch/heed/blob/5a10a00e02669d28edd6d11d35b617a9128189ca/heed/src/mdb/lmdb_ffi.rs#L46-L48).

```
thread '<unnamed>' panicked at library/core/src/panicking.rs:218:5:
unsafe precondition(s) violated: slice::from_raw_parts requires the pointer to be aligned and non-null, and the total size of the slice not to exceed `isize::MAX`
```

### DIFF
--- a/heed3/Cargo.toml
+++ b/heed3/Cargo.toml
@@ -26,9 +26,10 @@ synchronoise = "1.0.1"
 
 [dev-dependencies]
 argon2 = { version = "0.5.3", features = ["std"] }
-memchr = "2.7.4"
-serde = { version = "1.0.217", features = ["derive"] }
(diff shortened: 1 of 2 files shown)


hyc:
I have no idea at the moment. Multithreaded reads works perfectly well with encryption in OpenLDAP slapd.

Have you tried testing with the example crypto.lm in the LMDB source?

PR #349, opened 2025-11-05
fix link to examples

uniphil:
just a little link path fix, since i guess the example folder was moved


Kerollmops (approved):
Indeed, thank you 😊

PR #351, opened 2025-11-18
Add method to get a value from a key/value

Boog900:
# Pull Request

## What does this PR do?

Adds a method to get a value from a (key, value). Although on the surface this may seem useless, if you define the values comparison function to only take into account part of the value then you can use this method to do direct lookups of multi-map values.

This method would probably be less necessary if we exposed Cursors, which I would like, however this is a much smaller change and fits my use-case.

### DIFF
--- a/heed/src/cursor.rs
+++ b/heed/src/cursor.rs
@@ -145,6 +145,27 @@ impl<'txn> RoCursor<'txn> {
         }
     }
 
+    pub fn move_on_key_value(&mut self, key: &[u8], data: &[u8]) -> Result<bool> {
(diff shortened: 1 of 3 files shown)


Kerollmops:
Hey @Boog900 👋

Thank you for the change, however, I'm wondering if we could instead work on exposing the internal Cursor I use in the Iterators.

What do you think?

Boog900:
I would support that and it would allow me to do more things to make my code more efficient, I added this method as I didn't want to make too much of a change to the API and it was the least I needed to get my code working.

PR #352, opened 2025-11-18
fix `last` on iterators with dup values

Boog900:
# Pull Request

## What does this PR do?

From what I can tell in the `last` method for the iterators we are checking if the current key we are pointing at is equal to the last key so that if we have already gone through the iterator we return `None`. This doesn't work for dup tables though as the values may be different even if the key is the same (I added a test case that fails).

Although this PR removes this entirely, always returning a value for `last`, even if the iterator has already returned `None` from `next`. The main reason is this simpler and there is no requirement that once an iterator returns None that it wont return Some again: https://doc.rust-lang.org/std/iter/trait.Iterator.html#tymethod.next

Sorry if I missed something that makes this change invalid though!

### DIFF
--- a/heed/src/databases/database.rs
+++ b/heed/src/databases/database.rs
@@ -434,6 +434,7 @@ impl<KC, DC, C, CDUP> Database<KC, DC, C, CDUP> {
     /// drop(iter);
     ///
     /// let mut iter = db.get_duplicates(&wtxn, &68)?.expect("the key exists");
+    /// assert_eq!(iter.next().transpose()?, Some((68, 120)));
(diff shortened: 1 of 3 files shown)


Kerollmops (changes requested):
Thank you @Boog900 for this PR, it's a breaking change, and would it be possible to document the `last` method about this specific behavior, please?

Kerollmops:
Hey @Boog900 👋

I hope you're good. Do you think you can take a quick look at my last message, please?

Boog900:
Hi, sorry, forgot about this. Pushed some comments

PR #353, opened 2025-11-19
preserve cmp functions with `DatabaseOpenOptions::types`

Boog900:
# Pull Request

## What does this PR do?

Changes the  `DatabaseOpenOptions::types` method to preserve the compare functions on the database, currently it resets them to the default which can lead to some hard to debug issues if this is used after setting the cmp types.


Kerollmops (approved):
Ho wow! You are right! My bad on that 🤦‍♂️ Thank you @Boog900 😊

PR #362, opened 2026-04-13
Remove dependency on the deprecated cstr crate

millert:
In Rust 1.77.0 and above, c"abc" can be used in place of cstr!("abc").


Kerollmops (dismissed):
Ho! I learned something today! Thank you. Great PR! Merging.

EDIT: after you do a small cargo fmt 😅

Kerollmops (approved):
Thanks for the PR. Merging. Have a nice day 🌵

PR #365, opened 2026-04-16
Bump LMDB submodules to the latest commits

Kerollmops:
This PR bumps the two LMDB dependencies we use: mdb.master.nested-rtxns and mdb.master3. The former is a fork that we maintain at Meilisearch; the latter is the official one. A lot of fixes and improvements were shipped recently.

Capture d’écran 2026-04-16 à 18 25 18

### DIFF
--- a/heed/src/mdb/lmdb_error.rs
+++ b/heed/src/mdb/lmdb_error.rs
@@ -186,7 +186,7 @@ mod test {
 
     #[test]
     fn test_description() {
-        assert_eq!("Permission denied", Error::from_err_code(13).to_string());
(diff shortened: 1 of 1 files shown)


Kerollmops:
After tracking the CI issue, I found that there was likely an issue in a recent LMDB commit. You can read more about this [on ITS#10454](https://bugs.openldap.org/show_bug.cgi?id=10454). I proposed a patch there. I am waiting for Howard Chu's response to rebase our patched LMDB (mdb.master.nested-rtxns) and bump both versions.

PR #373, opened 2026-06-15
Introduce `static_write_txn`

Paraworker:
# Pull Request

## What does this PR do?
  - Adds `Env::static_write_txn` and `EncryptedEnv::static_write_txn`, mirroring the existing `static_read_txn` API. These constructors consume the env (`self`) and return an
   `RwTxn<'static>` whose lifetime is no longer tied to a borrowed `Env`, letting callers hold a write txn without keeping a separate `Env` reference alive.
  - Implementation in `heed/src/txn.rs` mirrors `RwTxn::static_read_txn`: begins an `mdb_txn_begin` write transaction and stores the env as `Cow::Owned` inside `RoTxnInner`
  so the env cannot be closed while the txn lives.
  - Updates the `write_txn` rustdoc on both `Env` and `EncryptedEnv` to cross-reference the new `static_write_txn`.

## PR checklist
- [x] Did you use any AI tool while implementing this PR (code, tests, docs, etc.)? If yes, disclose it in the PR description and describe what it was used for. AI usage is allowed when it is disclosed.
    - Used Claude Code to explore the repo structure and locate the existing `static_read_txn` implementation, then mirrored its shape for `static_write_txn` (env consumed by value, `Cow::Owned` env handle, `'static` lifetime on the returned txn). Final code and docs were reviewed by hand.

### DIFF
--- a/heed/src/txn.rs
+++ b/heed/src/txn.rs
@@ -279,6 +279,14 @@ pub struct RwTxn<'p> {
 
 impl<'p> RwTxn<'p> {
     pub(crate) fn new<T>(env: &'p Env<T>) -> Result<RwTxn<'p>> {
+        Self::begin(Cow::Borrowed(&env.inner))
+    }
+
+    pub(crate) fn static_write_txn<T>(env: Env<T>) -> Result<RwTxn<'static>> {
+        Self::begin(Cow::Owned(env.inner))
+    }
+
+    fn begin(env: Cow<'_, Arc<EnvInner>>) -> Result<RwTxn<'_>> {
         let mut txn: *mut ffi::MDB_txn = ptr::null_mut();
 
         unsafe {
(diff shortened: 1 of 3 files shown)


heed/src/txn.rs
```diff
@@ -279,6 +279,14 @@ pub struct RwTxn<'p> {
...
 impl<'p> RwTxn<'p> {
     pub(crate) fn new<T>(env: &'p Env<T>) -> Result<RwTxn<'p>> {
+        Self::begin(Cow::Borrowed(&env.inner))
+    }
+
+    pub(crate) fn static_write_txn<T>(env: Env<T>) -> Result<RwTxn<'static>> {
+        Self::begin(Cow::Owned(env.inner))
+    }
+
+    fn begin(env: Cow<'_, Arc<EnvInner>>) -> Result<RwTxn<'_>> {
```
Kerollmops, line 276: Would you mind changing a bit the naming and doing the same naming changes for the RoTxn struct, please?
- `static_read_txn` becomes `from_owned_env`
- `static_write_txn` becomes `from_owned_env`
- apply the same technique to provide a `Cow` as a parameter to the `begin` method for the `RoTxn` too.

Kerollmops (changes requested):
Hello @Paraworker,

Thanks for the changes. Looks good to me. I would just like you to clean a bit the codebase.

Thank you very much 🌵

heed/src/txn.rs
Paraworker: Changed as suggested.

PR #375, opened 2026-07-09
Add serde-wincode as an alternative to bincode

A-Manning:
## What does this PR do?
This PR adds `serde-wincode` as an alternative to bincode, which is considered deprecated. `serde-wincode` is intended to be a drop-in replacement for `bincode`'s `serde` compatibility.

## PR checklist
- [x] Did you use any AI tool while implementing this PR (code, tests, docs, etc.)? If yes, disclose it in the PR description and describe what it was used for. AI usage is allowed when it is disclosed.

### DIFF
--- a/heed-types/Cargo.toml
+++ b/heed-types/Cargo.toml
@@ -14,13 +14,15 @@ byteorder = "1.5.0"
 heed-traits = { version = "0.20.0", path = "../heed-traits" }
 serde = { version = "1.0.223", optional = true }
 serde_json = { version = "1.0.145", optional = true }
+serde-wincode = { version = "0.1.2", optional = true }
 rmp-serde = { version = "1.3.0", optional = true }
 
 [features]
-default = ["serde-bincode", "serde-json"]
-serde-bincode = ["serde", "bincode"]
-serde-json = ["serde", "serde_json"]
-serde-rmp = ["serde", "rmp-serde"]
+default = ["serde-json", "serde-wincode"]
+serde-bincode = ["dep:serde", "bincode"]
+serde-json = ["dep:serde", "serde_json"]
+serde-rmp = ["dep:serde", "rmp-serde"]
+serde-wincode = ["dep:serde", "dep:serde-wincode"]
 # serde_json features
 preserve_order = ["serde_json/preserve_order"]
 arbitrary_precision = ["serde_json/arbitrary_precision"]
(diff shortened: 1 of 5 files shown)


Kerollmops (changes requested):
Thanks again @A-Manning,

I am wondering if it wouldn't be preferable to not introduce another dependency to a crate in heed. I would rather prefer we mark the bincode codec types as deprecated so that we can remove them in a future version.

The best approach would be to create a new heed-bincode crate that exposes the current codecs and also mark them as deprecated (or not).

This way we can inform users that they should switch to this dedicated crate instead of using the `bincode` heed feature.

Then, you can, independently of me or the heed organization, introduce a new `heed-wincode` crate with the appropriate codecs.

Ho! And don't forget to remove the `remove-me` commit when you push on a branch and open a PR with it. It prevents the CI from running and the CI will already convert the code for heed and heed3.

Thanks for your help 🌵

A-Manning:
>I am wondering if it wouldn't be preferable to not introduce another dependency to a crate in heed. I would rather prefer we mark the bincode codec types as deprecated so that we can remove them in a future version.

The `bincode` codec types are marked as deprecated in this PR, and are no longer included in the default feature set.
The default feature set includes a dependency on `serde-wincode`, which is equivalent in functionality to the `bincode` crate. The number of direct dependencies with default features is unchanged, but there is 1 more if all features are enabled (although, it would be redundant to enable both `serde-bincode` and `serde-wincode`)

>The best approach would be to create a new heed-bincode crate that exposes the current codecs and also mark them as deprecated (or not).

I don't think this is worth setting up new crates. The `heed-bincode`, `heed-json`, `heed-rmp`, and `heed-wincode` crates would each be ~30-40 LoC. It makes more sense to keep them as optional features in `heed-types`.
Also worth pointing out that these codecs are specific to serde serializers, not serialization formats, and so should be named something like `heed-serde-bincode`, `heed-serde-json`, etc.

A-Manning:
On the topic of default features, I think it would be best not to enable `serde-json` and `serde-bincode` (or `serde-wincode`, after this PR) by default. Users should explicitly enable the codecs that they actually want. Pulling in `bincode` or `serde-json` as dependencies by default is a surprising and likely unwanted behavior. `heed` also has `serde` as a default feature, whereas most other crates make `serde` an optional, non-default feature.

I have preserved the current behavior in this PR, but IMO it would be better to have no default features in `heed-types` AND `heed`.

PR #376, opened 2026-07-09
RoTxn: impl AsMut/AsRef/Borrow/BorrowMut

A-Manning:
It is awkward to specify bounds in terms of the existing `Deref` impl, ie.

```
fn foo<'a, Tls>(rotxn: &RoTxn<'a, Tls>) where RoTxn<'a, Tls>: Deref<Target = RoTxn<'a>> {
    ...
}
```

with this PR, the bound is no longer necessary, since any `&RoTxn<'a, Tls>` can be converted to `&RoTxn<'a, AnyTls>`.

### DIFF
--- a/heed/src/txn.rs
+++ b/heed/src/txn.rs
@@ -1,4 +1,5 @@
-use std::borrow::Cow;
+use std::borrow::{Borrow, Cow};
+use std::convert::AsRef;
 use std::marker::PhantomData;
 use std::ops::Deref;
 use std::ptr::{self, NonNull};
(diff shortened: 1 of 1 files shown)


Kerollmops (commented):
Thanks @A-Manning,

I think your PR is good and important to simplify the use the `RoTxn` with an `AnyTls` parameter.


Would you mind deleting the `remove-me` commit, please? I think that's what's preventing the CI from running 🤔

A-Manning:
>Would you mind deleting the remove-me commit, please?

Removed, CI should pass now

